PR #2, opened 2017-05-17
Disabled remember_me checkbox by default.

LucaPipolo:
On the Asana ticket @sun mentioned that the plugin should already hide Remember Me checkbox as default behaviour:

> […] but shouldn't have been, because core-standards hides it by default – or should hide it by default.

Instead, checking the code, I found a comment that said just the opposite [`// Enable Remember Me by default.`]. In addition, this code seems not working since also if values are changed (from 1 to 0 and from TRUE to FALSE) the checkbox is still displayed.

I research a little bit and I found that we should:

1. manipulate the `remember` values of the `$args` array using the `login_form_defaults` hook for the login form displayed in the frontend;
2. use a PHP `preg_replace` (better than an inline CSS solution IMO) to clean the backend form in the `wp-login.php`. 

In this PR I tried to follow this two steps but since I'm not sure what the previous code did and why it was implemented (Different intended behaviour? Outdated code?) I don't change it yet, waiting a feedback first.

### DIFF
--- a/src/UserFrontend.php
+++ b/src/UserFrontend.php
@@ -61,6 +61,7 @@ class UserFrontend {
    */
   public static function login_form_defaults(array $args) {
     $args['value_remember'] = TRUE;
+    $args['remember'] = FALSE;
     return $args;
   }
 
@@ -87,6 +88,9 @@ body, html {
   background-size: auto;
   width: auto;
   height: 50px;
+}
+#loginform .forgetmenot {
+  display: none;
 }
 </style>
 


sun:
> I found a comment that said just the opposite `// Enable Remember Me by default.`

@LucaPipolo "Enabling" is not the opposite of "hiding".  The checkbox can be enabled by default AND hidden. The only desired change is to hide the checkbox (and forcing the value to be enabled).

sun (changes requested):
> I found a comment that said just the opposite `// Enable Remember Me by default.`

@LucaPipolo "Enabling" is not the opposite of "hiding".  The checkbox can be enabled by default AND hidden. The only desired change is to hide the checkbox (and forcing the value to be enabled).

src/UserFrontend.php
```diff
@@ -61,10 +64,26 @@ public static function login_head() {
...
     $args['value_remember'] = TRUE;
+    $args['remember'] = FALSE;
     return $args;
   }
 
   /**
+   * @implements login_form
+   */
+  public static function login_form() {
+    ob_start(__CLASS__ . '::removesForgetmenotCheckbox');
```
sun: When you start an output buffer, you also need to end and clean it.

Do we actually need the output buffer?  Setting the `remember` parameter should hide the checkbox already?

src/UserFrontend.php
```diff
@@ -15,7 +15,10 @@ public static function init() {
 
     // Enable Remember Me by default.
     add_action('login_head', __CLASS__ . '::login_head');
-    add_filter('login_form_defaults', __CLASS__ . '::login_form_defaults');
+    add_filter('login_form_defaults', __CLASS__ . '::login_form_defaults', 99);
```
sun: Why do we a need a higher/later priority than 10?

sun (changes requested):

fabianmarz:
Why do we want this behaviour for hiding the checkbox but have it enabled by default? What is the corresponding asana ticket?

src/UserFrontend.php
LucaPipolo: You're right. Sorry: forgot to fix this after some tests.

src/UserFrontend.php
LucaPipolo: As I already told before in the PR description, reading lots of documentation online, we need two different steps in order to hide the checkbox everywhere. The `remember` parameter hide the checkbox everywhere but not on the `wp-login.php` page. To hide it as well on the `wp-login.php` we need to use the `login_form` action as well.

About the output buffer is needed to alter the `$content` since, without it, the parameter is empty. `login_form` is an action and is called just before the remember me checkbox is output so this is why we need to start an output buffer to alter the content is coming next (see `wp-login.php:907`).  Before deciding to adopt this solution, I studied a couple of plugins that remove the checkbox and both of them use the same approach. 

Agree about the buffer end/clean. Where to do it in your opinion? At the end of the `removesForgetmenotCheckbox` function? 

On a separate topic: should, in your opinion, `removesForgetmenotCheckbox` be more consistent if called `removedRememberMeCheckbox`?

LucaPipolo:
@fabianmarz — [here](https://app.asana.com/0/53495308123589/331784584128184) there is the corresponding Asana ticket that is a follow-up of [this other one](https://app.asana.com/0/53495308123589/332829600085024).

If you're asking me why it should be done as default: I asked it to @sun (in the task comments) why this should be a default behaviour to share across all our projects that use `core-standards` plugin. In his reply, he mentioned:

> The idea is that we want to (1) enable it by default and (2) hide it by default on all sites, because it is almost always needless clutter for the average website user. So for the sake of good UX design, we choose a sane default instead.

and 

> Yes, the checkbox was visible, but shouldn't have been, because core-standards hides it by default – or should hide it by default.

src/UserFrontend.php
sun: Hiding/removing the checkbox entirely from the markup is a good idea for the frontend (which is achieved with the simple parameter).

For the backend login form, I think the solution can be more simple.  I'd simply load/output CSS to hide the form element in the frontend.

LucaPipolo: @sun — ok. I supposed, as said in the description, that remove the output element could be better that hide it via CSS. However, if you prefer this option: let's do that. 

Fixup commit on the way. 🙂

src/UserFrontend.php
```diff
@@ -61,13 +61,22 @@ public static function login_head() {
...
   /**
    * @implements login_enqueue_scripts
    */
   public static function login_enqueue_scripts() {
+    echo <<<EOD
+<style>
+#loginform .forgetmenot {
+  display: none;
+}
+</style>
```
sun: Wondering whether the style should be added to the existing default CSS below?

Existing sites with a custom `login.css` file in their theme would remain to have the remember_me checkbox visible then – which sounds appropriate to me.

sun (changes requested):

src/UserFrontend.php
LucaPipolo: Agree. I did it because I was not sure about the correct beahviour. 
Is not the existing CSS below printed only if a `login.css` file not exists?

```
if (file_exists(get_stylesheet_directory() . '/dist/styles/login.css')) {
…
} 
else {
  echo <<<EOD
<style>
…
}
```

So if I put the code inside that style we should, in SVC case, add same CSS rule also to `login.css`. Is it correct?

sun: Yes, that's correct, in case there is a login.css in the svc theme already.

sun: Did you forget to add the change in login.css?

LucaPipolo: @sun — nope… `login.css` is in the theme… in this case SVC but this is a PR for `wordpress-core-standards` plugin so I suppose that change to `login.css` should be applied as commit to the SVC repo, am I wrong? 

However in the SVC case it was already here in commit `df469492`.

sun (approved):
Please squash all commits into a single and use the following commit message:

Hide "remember me" form item in user login form (but keep it enabled) by default.

PR #3, opened 2017-06-01
Remove width and height attributes from SVG images

fabianmarz:
Wordpress is not generating any size metadata for svg files. So svg files will have a default of `width="1"` and `height="1"` when the image string is rendered.

This PR prevents the default width and height attributes.

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -46,6 +46,13 @@ class Plugin {
     // Allow SVG files in media library.
     add_filter('upload_mimes', __CLASS__ . '::upload_mime_types');
 
+    // Remove size attributes from SVG images, so they scale correctly without a CSS size reset.
+    add_filter('the_content', __CLASS__ . '::the_content');
+    add_filter('wp_get_attachment_metadata', __CLASS__ . '::removeSvgSizeAttributes', 10, 2);
+    add_filter('post_thumbnail_html', __CLASS__ . '::removeSvgSizeAttributes', 10, 3);
+    // Limit to two arguments since the third parameter is the image caption.
+    add_filter('image_send_to_editor', __CLASS__ . '::removeSvgSizeAttributes', 10, 2);
+
     // Set proper From header for all emails.
     // Remove "[$blogname]" prefix in email subjects of user account mails.
     add_filter('wp_mail', __NAMESPACE__ . '\Mail::wp_mail');
@@ -83,6 +90,46 @@ class Plugin {
   public static function upload_mime_types(array $mimes) {
     $mimes['svg'] = 'image/svg+xml';
     return $mimes;
+  }
+
+  /**
+   * Removes SVG image size attributes from the content.
+   *
+   * @implements the_content
+   */
+  public static function the_content($content) {
+    $content = preg_replace_callback('@<img [^>]*src=[^>]+?.svg[^>]*>@', function ($matches) {
+      return static::removeSizeAttributes($matches[0]);
+    }, $content);
+    return $content;
+  }
+
+  /**
+   * Removes size attributes from SVG images.
+   *
+   * @implements wp_get_attachment_metadata
+   * @implements image_send_to_editor
+   * @implements post_thumbnail_html
+   */
+  public static function removeSvgSizeAttributes($data, $post_id, $attachment_id = NULL) {
+    if (!$data || get_post_mime_type($attachment_id ?? $post_id) !== 'image/svg+xml') {
+      return $data;
+    }
+    // For wp_get_attachment_metadata, $data is an array.
+    if (is_array($data)) {
+      unset($data['width'], $data['height']);
+    }
+    else {
+      $data = static::removeSizeAttributes($data);
+    }
+    return $data;
+  }
+
+  /**
+   * Removes size attributes from the given HTML string.
+   */
+  public static function removeSizeAttributes($html) {
+    return preg_replace('@(?:width|height)="1"\s+@', '', $html);
   }
 
   /**


src/Plugin.php
```diff
@@ -86,6 +90,45 @@ public static function upload_mime_types(array $mimes) {
   }
 
   /**
+   * Removes svg width and height attributes from the content.
+   *
+   * @implements the_content
+   */
+  public static function the_content($content) {
+    $content = preg_replace_callback('@<img [^>]*src=.+?.svg[^>]*>@', function ($matches) {
```
sun: The last remaining `.+?` needs to be replaced with `[^>]+?`

src/Admin.php
```diff
@@ -32,6 +32,7 @@ public static function init() {
     // Exposes SVG images in media library.
     add_filter('wp_prepare_attachment_for_js', __CLASS__ . '::wp_prepare_attachment_for_js');
     add_action('admin_head', __CLASS__ . '::admin_head');
+    add_filter('image_send_to_editor', __NAMESPACE__ . '\Plugin::removeSvgWidthHeightAttributes', 10, 2);
```
sun: 1. The added line should be separated with a blank line and an own inline comment that explains what it does, because it does not belong to exposing SVG images in the media library.

1. The registration of this filter callback can be moved to the other registered filters in `Plugin::init()`, so that everything is in one place.  (When we talked about it, I assumed there to be more than just the registration.)

src/Plugin.php
```diff
@@ -86,6 +90,45 @@ public static function upload_mime_types(array $mimes) {
...
+      return preg_replace('@((?:width|height)=.+?\s+)@', '', $matches[0]);
```
sun: 1. The intention of `.+?\s+` seems to be `[^\s]+\s+`

1. The outer parenthesis are not necessary and should be removed, because you do not use the capturing group.

1. This will also remove the width and height of images for which it was set manually and explicitly (e.g., logos in an article text). Didn't you state that the lack of image dimensions can be determined by `width="1" height="1"`?

src/Plugin.php
```diff
@@ -86,6 +90,45 @@ public static function upload_mime_types(array $mimes) {
...
+    }, $content);
+    return $content;
+  }
+
+  /**
+   * Removes svg width and height attributes when the image is received via a function.
```
sun: 1. SVG is an acronym and should thus always be written in all uppercase letters. (also elsewhere in this PR)

1. What do you mean with _"when the image is received via a function"_ ? That's a bit broad. What kind of function?

src/Plugin.php
```diff
@@ -59,6 +59,10 @@ public static function init() {
     add_action('rss2_item', __NAMESPACE__ . '\Feed::rss2_item');
 
     UserFrontend::init();
+
+    add_filter('the_content', __CLASS__ . '::the_content');
+    add_filter('wp_get_attachment_metadata', __CLASS__ . '::wp_get_attachment_metadata', 10, 2);
+    add_filter('post_thumbnail_html', __CLASS__ . '::removeSvgWidthHeightAttributes', 10, 3);
```
sun: 1. There should be an inline comment above this block of filter registrations, explaining the overall feature (i.e., all filters combined).

1. Why is the third callback not named `post_thumbnail_html`?

sun (changes requested):

src/Plugin.php
fabianmarz: Why is it needed to be replaced? I want to match everything in between the src attribute which ends on .svg. Any specific reason why this could not be achieved in this way? In addition with `.+?` the regex takes only 348 steps while the regex with `[^>]+?` takes 427. Not that significant. Just pointed out.

src/Plugin.php
fabianmarz: 1. 👍 
2. It's not named like the filter, because it's also used from within the Admin class. So I thought it would make more sense to have a more specific method name.

src/Plugin.php
fabianmarz: 1. Yes, also here, why can't it be achieved like it is currently?
2. 👍 Thanks for pointing out! 
3. This is partially possible. For example in SCRM we set the default value for width/height to 128 if it was not present. So those won't be stripped then.

src/Plugin.php
fabianmarz: 1. 👍 
2. E.g. `get_attachment_image()`, `get_post_thumbnail()` etc. anywhere where the metadata for this image is requested. Probably it's better to rewrite the comment and name *when the meta data for the attachment is received* or something.

src/Plugin.php
sun: Because `.+?.svg` causes it to also match this (entire string):

```
<img src="foo.png">whatever<img src="bar.svg">
```
---
Lastly, I also just noticed that you did not escape the dot before the `svg` file extension, so it actually matches any character instead of only a period.

src/Plugin.php
sun: 1. Doesn't make a practical difference in this case here, as far as I can see.

1. 👍

1. Can we fix the SCRM database (changing 128 into 1) and make this code look for the explicit case of `width="1" height="1"`?

src/Plugin.php
sun: 1. Thanks – forgot to mention that class names and method names never use all-uppercase acronyms, so it stays "Svg" there.

src/Plugin.php
```diff
@@ -45,6 +45,11 @@ public static function init() {
 
     // Allow SVG files in media library.
     add_filter('upload_mimes', __CLASS__ . '::upload_mime_types');
+    // Register filters to remove the width and height attributes for SVG files.
```
sun: 1. Still missing a newline to previous chunk (see chunks before and after this for examples).

1. The comment is too technical. It should be a higher-level description here, such as _"Remove size attributes from SVG images, so they scale correctly."_

src/Plugin.php
```diff
@@ -59,6 +64,7 @@ public static function init() {
     add_action('rss2_item', __NAMESPACE__ . '\Feed::rss2_item');
 
     UserFrontend::init();
+
```
sun: Straw newline

src/Plugin.php
```diff
@@ -86,6 +92,46 @@ public static function upload_mime_types(array $mimes) {
...
+  public static function the_content($content) {
+    $content = preg_replace_callback('@<img [^>]*src=.+?.svg[^>]*>@', function ($matches) {
+      return preg_replace('@(?:width|height)=.+?\s+@', '', $matches[0]);
+    }, $content);
+    return $content;
+  }
+
+  /**
+   * Removes SVG width and height attributes when the metadata for the attachment
+   * is received. E.g when calling get_attachment_image() or get_post_thumbnail().
```
sun: The PHPDoc summary should not wrap but be on a single line, and thus needs to be a short and concise as possible; e.g., _"Removes size attributes from SVG images when requesting metadata."_

You may reference the two functions via `@see` instead — although that context might be sufficiently clear by the `@implements` already? (not sure)

src/Plugin.php
fabianmarz: But what's about e.g.: [DOMDocument](http://php.net/manual/de/class.domdocument.php) and [saveXML](http://php.net/manual/de/domdocument.savexml.php)? They also using acronyms in uppercase for methods/classnames.

src/Plugin.php
fabianmarz: Oh dear, thanks for pointing out!

src/Plugin.php
fabianmarz: 🙈

src/Plugin.php
fabianmarz: Yes, you're right. I'll shorten this and drop the example functions as this is clear by the `@implements` doc.

src/Plugin.php
sun: https://www.drupal.org/node/1627350#comment-6103372

src/Plugin.php
fabianmarz: 3. Yes we should change this! Will need an extensive regex which will take long to finish when called via search-replace but this would be the best solution.

src/Plugin.php
sun: Still here?

src/Plugin.php
```diff
@@ -86,6 +93,45 @@ public static function upload_mime_types(array $mimes) {
...
+
+  /**
+   * Removes size attributes from SVG images.
+   *
+   * @implements wp_get_attachment_metadata
+   * @implements image_send_to_editor
+   * @implements post_thumbnail_html
+   */
+  public static function removeSvgSizeAttributes($data, $post_id, $attachment_id = NULL) {
+    if ($data && get_post_mime_type($attachment_id ?? $post_id) === 'image/svg+xml') {
```
sun: For [`image_send_to_editor`](https://developer.wordpress.org/reference/hooks/image_send_to_editor/), the 3rd argument is the image caption, which is not an attachment ID.

src/Plugin.php
```diff
@@ -86,6 +93,45 @@ public static function upload_mime_types(array $mimes) {
...
+    }
+    return $data;
+  }
+
+  /**
+   * Removes size attributes from a given string or array.
+   */
+  public static function removeSizeAttributes($data) {
+    if (is_array($data)) {
+      unset($data['width'], $data['height']);
```
sun: What you're doing here is called "function argument overloading" and generally discouraged, because the lack of type safety makes the code much harder to read and understand.  Two separate functions, of which one takes an array, and the other takes a string would be better.  (Given that the array only happens in a single case, you do not even need a helper function.)

If you want to keep the current code, we should at least document that the array only happens for the hook `wp_get_attachment_metadata`.

sun (changes requested):

src/Plugin.php
fabianmarz: Yes, this is why I'm registering the filter with 2 possible arguments only to not getting the caption in there:
`add_filter('image_send_to_editor', __CLASS__ . '::removeSvgSizeAttributes', 10, 2);`

src/Plugin.php
fabianmarz: I agree for bit more complicated readability. But isn't it better because it avoids repetition? If we change it to two separate functions we have to distinguish upfront which function to call which I wanted to avoid extensive nesting. Optionally we could do an early return. Also I'd like to stick with a helper function as the regex is used 2 times.

```php
public static function removeSvgSizeAttributes($data, $post_id, $attachment_id = NULL) {
  if ($data && get_post_mime_type($attachment_id ?? $post_id) === 'image/svg+xml') {
    if (is_array($data)) {
      unset($data['width'], $data['height']);
    }
    else {
      $data = static::removeSizeAttributes($data);
    }
  }
  return $data;
}
```
or
```php
public static function removeSvgSizeAttributes($data, $post_id, $attachment_id = NULL) {
  if (!$data || get_post_mime_type($attachment_id ?? $post_id) !== 'image/svg+xml') {
    return $data;
  }
  if (is_array($data)) {
    unset($data['width'], $data['height']);
  }
  else {
    $data = static::removeSizeAttributes($data);
  }
  return $data;
}
```

I don't like that both solutions much which I'd rather stick with the overladed function but add a comment. What do you think?

sun: Both of these variants look better to me than the overloaded function.  Regardless of which you choose, there should be a inline comment which explains that the array case is limited to `wp_get_attachment_metadata`.

src/Plugin.php
sun: Hm, not exactly obvious or trivial. Might also be worth a short inline comment.

src/Plugin.php
sun: Still here.

src/Plugin.php
```diff
@@ -86,6 +93,46 @@ public static function upload_mime_types(array $mimes) {
...
+   *
+   * @implements wp_get_attachment_metadata
+   * @implements image_send_to_editor
+   * @implements post_thumbnail_html
+   */
+  public static function removeSvgSizeAttributes($data, $post_id, $attachment_id = NULL) {
+    if (!$data || get_post_mime_type($attachment_id ?? $post_id) !== 'image/svg+xml') {
+      return $data;
+    }
+    // The array case is limited to wp_get_attachment_metadata.
```
```
// For wp_get_attachment_metadata, $data is an array.
```

src/Plugin.php
```diff
@@ -86,6 +93,46 @@ public static function upload_mime_types(array $mimes) {
...
+      unset($data['width'], $data['height']);
+    }
+    else {
+      $data = static::removeSizeAttributes($data);
+    }
+    return $data;
+  }
+
+  /**
+   * Removes size attributes from the given html string.
```
sun: HTML

sun:
You may squash into a single commit.  Don't forget to remove the straw newline at the end of the init() method.

src/Plugin.php
fabianmarz: Even better, thanks!

sun (approved):
Thanks for improving core standards!

sun (changes requested):
That said, the commit message needs polishing.

Most importantly, it should state the "actual" reason, the root cause, the original trigger that led to the idea of this change.

A description and explanation of the technical change may follow in the commit message description, after the summary.  In there, you would explain why removing these size attributes is a good idea, and why sizing cannot be solved with pure CSS.

Lastly, SVG is an acronym.  Acronyms are always written in all-uppercase in all natural text; i.e., comments, documentation, and commit messages.

fabianmarz:
@sun can you please have a final look of my force pushed commit?

sun:
```diff
 Fixed SVG files do not scale properly due to size attribute defaults.

-Wordpress do not generate SVG size metadata which will set the html width and height attributes to a default of 1.
+WordPress does not generate SVG size metadata, which causes the HTML width and height attributes to be set to a default of 1.
 To prevent false scaling, the sizes should be set via CSS.
```

Not sure I understand the last sentence – doesn't that imply that this:
```css
[src*='.svg'] {
  width: auto;
  height: auto;
}
```
would fix the problem equally well with pure CSS?

It would be good to document (in code or at least commit message) why that solution would be inferior to the chosen one.

fabianmarz:
I had a talk with @timogauweiler told him the possible options. As a conclusion we decided to drop the size attributes completely and do the styling via CSS.

```
[src*='.svg'] {
  width: auto;
  height: auto;
}
```
This will solve the problem, yes! This enhancement of the core-standards simply prevents the size attributes, so they won't appear in the rendered strings. The styling for the various sizes has to be done per project.

sun:
Okay, in that case, you intentionally opted for a more complex solution in order to not see the attributes in the output (cosmetics) and to not have to do the CSS reset – which should be documented exactly like this (in code and/or commit message), so that future generations are able to decide whether this is still the best/desired solution or not.

src/Plugin.php
```diff
@@ -46,6 +46,13 @@ public static function init() {
     // Allow SVG files in media library.
     add_filter('upload_mimes', __CLASS__ . '::upload_mime_types');
 
+    // Remove size attributes from SVG images, so they scale correctly without a CSS size reset.
```
fabianmarz, line 49: @sun - I changed the comment here and adjusted the commit description. Thanks for reviewing!

sun (approved):

PR #4, opened 2017-07-19
Fixed Facebook embeds keep within wrapper borders.

juliquiron:
Problem
* After updating WordPress Core recently, Facebook content embedded through a oEmbed URL exceeds the available space (width) of the layout / main content area on one of our websites.

Details
* The embed code of `wp-includes/class-oembed.php` forces the the `maxwidth` parameter as integer for the oembed endpoint.
* This works with all platforms as CSS `max-width` with an CSS line setting `width: 100%`.
* But for Facebook the output is `width: 750px`, which is forced via JavaScript.

Proposed solution
1. Use `oembed_fetch_url` filter to modify the query params overriding `max-width` with a '100%' string in case the content is from facebook.

Notes
* The oEmbed cache is stored in `wp_postmeta` with keys `_oembed_RANDOMHASH`.

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -65,6 +65,9 @@ class Plugin {
     // Add teaser image to RSS feeds.
     add_action('rss2_item', __NAMESPACE__ . '\Feed::rss2_item');
 
+    // Fix Facebook embeds to keep within wrapper borders.
+    add_filter('oembed_fetch_url', __CLASS__ . '::oembed_fetch_url', 10, 3);
+
     UserFrontend::init();
   }
 
@@ -151,6 +154,18 @@ class Plugin {
    */
   public static function getBasePath() {
     return dirname(__DIR__);
+  }
+
+  /**
+   * Fix Facebook embeds to keep within wrapper borders.
+   *
+   * @implements oembed_fetch_url
+   */
+  public static function oembed_fetch_url($provider, $url, $args) {
+    if (strpos($url, 'facebook.com')) {
+      $provider = add_query_arg( 'maxwidth', '100%', $provider );
+    }
+    return $provider;
   }
 
 }
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -65,7 +65,7 @@ class Plugin {
     // Add teaser image to RSS feeds.
     add_action('rss2_item', __NAMESPACE__ . '\Feed::rss2_item');
 
-    // Fix Facebook embeds to keep within wrapper borders.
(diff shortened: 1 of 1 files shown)


src/Plugin.php
```diff
@@ -65,6 +65,9 @@ public static function init() {
     // Add teaser image to RSS feeds.
     add_action('rss2_item', __NAMESPACE__ . '\Feed::rss2_item');
 
+    // Fix Facebook embeds to keep within wrapper borders.
+    add_filter('oembed_fetch_url', __CLASS__ . '::oembed_fetch_url', 10, 3);
```
sun: 1. There's a oEmbed section at the beginning of `init()` – which adds support for Facebook as an oEmbed provider in the first place.  (WordPress Core does not support Facebook natively.)

1. The comment could be more clear – _"Ensure that Facebook embeds do not exceed available content width."_

src/Plugin.php
```diff
@@ -153,4 +156,16 @@ public static function getBasePath() {
...
   }
 
+  /**
+   * Fix Facebook embeds to keep within wrapper borders.
+   *
+   * @implements oembed_fetch_url
+   */
+  public static function oembed_fetch_url($provider, $url, $args) {
+    if (strpos($url, 'facebook.com')) {
+      $provider = add_query_arg( 'maxwidth', '100%', $provider );
```
sun: Superfluous spaces before and after the function argument list.

sun (changes requested):

src/Plugin.php
juliquiron: About 1, it's true but makes no sense that this is before the `is_admin()` exit point. Maybe we can move some of this filters after the `is_admin()` return? Otherwise, we can move this up, but we will be adding the filter in admin section where not necessary.

sun: Not sure whether that is true.  oEmbed content might actually be executed in the admin as well – e.g., when generating previews of content and possibly also for the rich-text editor preview.   Due to that possibility, the hooks are also registered in the admin.

sun (commented):
Wouldn't it make sense to place the new `oembed_fetch_url` before the existing `embed_oembed_html`, because fetching comes before output, and because the new hook is actually related to the very first oembed integration hook (also about Facebook) and less related to the second?

Also, the PHPDoc summary line still has the old comment?

juliquiron:
Updated in the order `proviers` -> `fetch_url` -> `oembed_html`. The comment is the new one :)

sun (approved):

PR #7, opened 2017-08-10
Changed wp_mail sender domain to be generated from wordpress internal function.

fabianmarz:
Due to a security vulnerability the email sender domain should not be generated from `$_SERVER['SERVER_NAME']` but use an internal WP function instead.

- https://exploitbox.io/vuln/WordPress-Exploit-4-7-Unauth-Password-Reset-0day-CVE-2017-8295.html
- https://core.trac.wordpress.org/ticket/25239#comment:91

### DIFF
--- a/src/Mail.php
+++ b/src/Mail.php
@@ -19,9 +19,7 @@ class Mail {
     $from_name = get_bloginfo('name');
     $from_email = get_option('admin_email');
     // Ensure that emails from staging site instances can still be identified.
-    if (isset($_SERVER['SERVER_NAME'])) {
-      $from_email = strtok($from_email, '@') . '@' . str_replace('www.', '', $_SERVER['SERVER_NAME']);
-    }
+    $from_email = strtok($from_email, '@') . '@' . str_replace('www.', '', parse_url(network_home_url(), PHP_URL_HOST));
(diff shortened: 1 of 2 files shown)


juliquiron (approved):
I see it totally correct. Maybe I will keep the comment line because the functionality is the same -just better implemented-, right?

fabianmarz:
Yes. You're right. Put the comment line back in.

src/Mail.php
```diff
@@ -19,9 +19,7 @@ public static function wp_mail(array $message) {
     $from_name = get_bloginfo('name');
     $from_email = get_option('admin_email');
     // Ensure that emails from staging site instances can still be identified.
-    if (isset($_SERVER['SERVER_NAME'])) {
-      $from_email = strtok($from_email, '@') . '@' . str_replace('www.', '', $_SERVER['SERVER_NAME']);
-    }
+    $from_email = strtok($from_email, '@') . '@' . str_replace('www.', '', parse_url(network_home_url(), PHP_URL_HOST));
```
sun, line 22: We should use the regular `site_url()` here, as `network_home_url()` returns the domain/URL of the operating network and not the current site itself.

PR #9, opened 2017-09-13
Fixed user logged out when closing the web browser.

juliquiron:
Problem:
- The users are logged out when closed the web browser

Details:
- On this commit, the checkbox was removed to activate the feature by default: https://github.com/netzstrategen/wordpress-core-standards/pull/2/files
- But as we can see there this is removing all the HTML element: https://core.trac.wordpress.org/browser/tags/4.8/src/wp-includes/general-template.php#L482
- When the login form is processed if the form element doesn't exist no value is taken: https://core.trac.wordpress.org/browser/tags/4.8/src/wp-login.php#L891

Proposed solution:
- Set back the form element and hide it via CSS

### DIFF
--- a/src/UserFrontend.php
+++ b/src/UserFrontend.php
@@ -60,8 +60,20 @@ class UserFrontend {
    * @implements login_form_defaults
    */
   public static function login_form_defaults(array $args) {
+    if (file_exists(get_stylesheet_directory() . '/dist/styles/login.css')) {
+      wp_enqueue_style('theme/login', get_stylesheet_directory_uri() . '/dist/styles/login.css');
+    }
+    else {
+      echo <<<EOD
+<style>
+#loginform .login-remember {
+  display: none;
+}
+</style
+
+EOD;
+    }
     $args['value_remember'] = TRUE;
-    $args['remember'] = FALSE;
     return $args;
   }
 
--- a/src/UserFrontend.php
+++ b/src/UserFrontend.php
@@ -60,19 +60,7 @@ class UserFrontend {
    * @implements login_form_defaults
    */
   public static function login_form_defaults(array $args) {
-    if (file_exists(get_stylesheet_directory() . '/dist/styles/login.css')) {
-      wp_enqueue_style('theme/login', get_stylesheet_directory_uri() . '/dist/styles/login.css');
-    }
-    else {
-      echo <<<EOD
-<style>
-#loginform .login-remember {
-  display: none;
-}
-</style
-
-EOD;
-    }
+    wp_enqueue_style('core-standards/login', Plugin::getBaseUrl() . '/assets/user-login-form.css');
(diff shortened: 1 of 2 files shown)


LucaPipolo (approved):
Looks good to me, thanks! 👍 

I also tested this changes on other websites, like SVC: all works properly.

src/UserFrontend.php
```diff
@@ -60,8 +60,20 @@ public static function login_head() {
    * @implements login_form_defaults
    */
   public static function login_form_defaults(array $args) {
+    if (file_exists(get_stylesheet_directory() . '/dist/styles/login.css')) {
+      wp_enqueue_style('theme/login', get_stylesheet_directory_uri() . '/dist/styles/login.css');
```
sun: This happens in line 85 already: https://github.com/netzstrategen/wordpress-core-standards/pull/9/files#diff-a4a15f99f66b9d0b641ea2f50c52c379R85

src/UserFrontend.php
```diff
@@ -60,8 +60,20 @@ public static function login_head() {
...
+    }
+    else {
+      echo <<<EOD
+<style>
+#loginform .login-remember {
+  display: none;
```
sun: This happens in line 105 already: https://github.com/netzstrategen/wordpress-core-standards/pull/9/files#diff-a4a15f99f66b9d0b641ea2f50c52c379R105

sun (changes requested):

src/UserFrontend.php
juliquiron: The selector is different because on line 105 applies on the login page.

juliquiron:
Thanks for your comments and testing.

@sun agree with it, I considered it and my points are:
- Each case applies to differents context. The new code applies to the form whatever is displayed, the old one only on the login page.
- The inline CSS is different, is only for one line and the selector isn't in the older piece of CSS, so cannot be reused.

Based on this a solution to not repeat code could be:
- A new method with  the condicional style inclusion paramatrizing the fallback inline style calladed in both existing methods. Makes sense to you?

sun:
Thanks for the clarification, these are good points.

In that case, instead of inline CSS, we should use a regular CSS file in the plugin that is included (enqueued) in the regular way.  If a theme wants to override the contained styles, it does so as usual; i.e., by dequeueing the plugin CSS or by simply overriding the styles in its own CSS.

The integration for `/wp-login.php` needs the special implementation to include the inline CSS, because there is no formal integration point for extensions on that page.

A simple `assets/user-login-form.css` file in the plugin is sufficient for this single rule, no need for a frontend build pipeline / gulp.

juliquiron:
+1 to the suggestion, so fixup pushed.

Can you check it again, please?

sun (approved):

PR #10, opened 2017-09-25
Fixed MCE editor sets unwanted rel attribute values for links that open in new window.

juanlopez4691:


src/Admin.php
```diff
@@ -32,6 +32,9 @@ public static function init() {
     // Exposes SVG images in media library.
     add_filter('wp_prepare_attachment_for_js', __CLASS__ . '::wp_prepare_attachment_for_js');
     add_action('admin_head', __CLASS__ . '::admin_head');
+
+    // Stops MCE editor setting unwanted rel attribute values for links that open in new window.
+    add_filter('tiny_mce_before_init', __CLASS__ . '::tinymce_allow_unsafe_link_target');
```
sun: The method name should be equal to the hook name, unless there is a reason to use a different name.

src/Admin.php
```diff
@@ -32,6 +32,9 @@ public static function init() {
     // Exposes SVG images in media library.
     add_filter('wp_prepare_attachment_for_js', __CLASS__ . '::wp_prepare_attachment_for_js');
     add_action('admin_head', __CLASS__ . '::admin_head');
+
+    // Stops MCE editor setting unwanted rel attribute values for links that open in new window.
```
sun:
```diff
-    // Stops MCE editor setting unwanted rel attribute values for links that open in new window.
+    // Prevents TinyMCE from setting unwanted rel attribute for links opening in new window.
```

Let's also move this new filter in between the sample_permalink and the SVG images, so that the filters are registered in context and logical execution order.

src/Admin.php
```diff
@@ -243,4 +246,17 @@ public static function admin_head() {
...
   }
 
+  /**
+   * Stops MCE editor setting attribute rel="noopener noreferrer"
+   * to links that opens in new window (target="_blank").
+   *
+   * @param array $mceInit MCD Editor configuration settings
+   *
+   * @return array
+   */
```
sun: 1. The PHPDoc summary does not wrap.  You can use the current text as PHPDoc description and add a new summary (consider to copy the single-line inline comment from init()).

1. The name of the editor is TinyMCE.

1. The parameter description should be on an own line. See `get_user_option_user_settings()` in this file.

1. Instead of manually specifying the hook implementation's parameters and return value, you should refer to the documented filter instead by stating `@implements`.  See `hidden_meta_boxes()` in this file.

1. In case there is WP Core issue relating to this bug/behavior, it would be great to link to it with a `@see` directive at the end of the PHPDoc.

src/Admin.php
```diff
@@ -243,4 +246,17 @@ public static function admin_head() {
...
+  public static function tinymce_allow_unsafe_link_target($mceInit) {
```
sun: We can type-hint the argument as `array`

sun (changes requested):

juanlopez4691:
Just pushed a fixup. Review, please?

src/Admin.php
```diff
@@ -209,6 +212,20 @@ public static function get_sample_permalink_html($html, $post_id, $new_title, $n
   }
 
   /**
+   * Prevents TinyMCE editor setting unwanted rel attribute values for links.
+   *
+   * TinyMCE sets attribute rel="noopener noreferrer" for links opening in new
+   * window (target="_blank"). This would prevent affiliates from correctly
+   * tracking those links.
```
sun:
```diff
-This would prevent affiliates from correctly tracking those links.
+This can prevent affiliate marketing tools from correctly tracking such links (e.g. in sponsored posts), potentially causing less revenue, so we disable the behavior by default.
```

src/Admin.php
```diff
@@ -209,6 +212,20 @@ public static function get_sample_permalink_html($html, $post_id, $new_title, $n
...
+   *
+   * @implements tiny_mce_before_init
+   */
```
sun: Let's add a pointer to the docs here:
```
@see https://www.tinymce.com/docs/configure/content-filtering/#allow_unsafe_link_target
```

sun (changes requested):

juanlopez4691:
@sun Pushed new fixup with last suggested corrections.

src/Admin.php
```diff
@@ -209,6 +212,22 @@ public static function get_sample_permalink_html($html, $post_id, $new_title, $n
...
+   *
+   * TinyMCE sets attribute rel="noopener noreferrer" for links opening in new
+   * window (target="_blank"). This can prevent affiliate marketing tools from
+   * correctly tracking such links (e.g. in sponsored posts), potentially causing
+   * less revenue, so we disable the behavior by default.
+   *
+   * @see https://www.tinymce.com/docs/configure/content-filtering/#allow_unsafe_link_target
+   * @implements tiny_mce_before_init
+   */
+   public static function tiny_mce_before_init(array $mceInit) {
```
sun: Wrong indentation here (3 spaces)

sun (changes requested):
Almost done!

src/Admin.php
juanlopez4691: Dang! Can't believe I missed that!

juanlopez4691:
Just pushed a new fixup. Can you review please?

sun (approved):

sun:
@juanlopez4691 Adjusted the commit message, please review.

PR #11, opened 2017-10-24
Added CLI commands for exporting/importing users.

sun:
`wp user export-csv` exports all users including core meta data and (hashed) passwords into a CSV file.

`wp user import-csv-raw` is a wrapper around the native `import-csv` subcommand that allows to import all data from a CSV file including already hashed passwords.

For this synchronization to work, the hash salts in wp-config.php need to be identical.

Warning: Install the plugin disable-emails to ensure that no emails are sent out during import.

Known issues:

- export-csv does not export all existing meta fields, because maintaining a consistent CSV column order across all user records is non-trivial.

- export-csv does not support any options to select/filter/limit the users to export.

- import-csv-raw does not support all options of import-csv.


src/Commands/UserImport.php
```diff
@@ -0,0 +1,52 @@
...
+   */
+  public function __invoke(array $args, array $options) {
+    $filename = $args[0];
+
+    if (isset($options['send-email'])) {
+      \WP_CLI::error('The option --send-email is not supported as it would send hashed passwords in emails to users.', TRUE);
+      exit;
+    }
+
+    $GLOBALS['wp_hasher'] = new class {
```
fabianmarz, line 36: What is this used for? Is this a built in WP global?

core-standards.php
```diff
@@ -39,3 +39,14 @@ function classloader($class) {
...
 add_action('init', __NAMESPACE__ . '\Plugin::init', 20);
 add_action('admin_init', __NAMESPACE__ . '\Admin::init');
+
+if (defined('WP_CLI') && WP_CLI) {
+  // Subcommands that extend an existing namespace wrongly replace the whole
+  // namespace, so each subcommand needs to be registered separately.
+  // @see https://make.wordpress.org/cli/handbook/commands-cookbook/#required-registration-arguments
+  // @see https://github.com/wp-cli/wp-cli/issues/1767
+  // @see https://github.com/wp-cli/wp-cli/issues/1795
+  // @see https://github.com/wp-cli/wp-cli/pull/1877
```
fabianmarz: Sorry, I don't really get the mentioned issues here. The issues are telling, that the subcommands are not listed when running `$ wp help`, which was fixed in v0.19.0. So why did you mention them here?

Also this paragraph can be found in the cookbook:
>All commands can be registered to their own top-level namespace (e.g. wp foo), or as subcommands to an existing namespace (e.g. wp core foo). For the latter, simply include the existing namespace as a part of the command definition.

Or is this working for parent commands but not already existing subcommands?

fabianmarz (changes requested):

src/Commands/UserImport.php
sun:
```suggestion
    // Inject a stub for the password hashing service of WordPress Core
    // to not re-hash the password on saving.
    $GLOBALS['wp_hasher'] = new class {
```

src/Commands/UserImport.php
```diff
@@ -0,0 +1,52 @@
...
+
+namespace Netzstrategen\CoreStandards\Commands;
+
+class UserImport extends \WP_CLI_Command {
+
+  /**
+   * Import users from a CSV file with no password hashing.
+   *
+   * Implement the filter hook 'user_contactmethods' to control which user meta
+   * fields are imported.
```
sun:
```suggestion
   * Note: You can implement the filter hook 'user_contactmethods' to control
   * which user meta fields are imported.
```

core-standards.php
sun: This comment was in relation to extending a namespace like this:
```php
  \WP_CLI::add_command('user', __NAMESPACE__ . '\Commands\User');
```
But you're right, the documentation has been updated in the meantime and suggests to register subcommands as already done in this PR, so the comments can be removed.

sun:
```suggestion
```

src/Commands/UserImport.php
```diff
@@ -0,0 +1,52 @@
...
+    // add_filter('user_contactmethods', function ($user) {
+    //   static $keys;
+    //   global $wpdb;
+    //   if (!isset($keys)) {
+    //     $keys = $wpdb->query("");
+    //   }
+    // });
+
+    \WP_CLI::run_command(array_merge(['user', 'import-csv'], $args), $options);
+    // \User_Command::import_csv($args, $options);
```
sun:
```suggestion
```

PR #12, opened 2017-10-25
Disable inbound pingbacks by default

juanlopez4691:


juanlopez4691:
Related Asana task: [Disable pingbacks in WordPress](https://app.asana.com/0/74314752394866/457006025051204/f)

src/Plugin.php
```diff
@@ -71,6 +71,9 @@ public static function init() {
     // Add teaser image to RSS feeds.
     add_action('rss2_item', __NAMESPACE__ . '\Feed::rss2_item');
 
+    // Disable pingbacks.
+    add_action('pre_ping', 'pre_ping');
```
sun: https://developer.wordpress.org/reference/hooks/pre_ping/ is for outbound pings to other sites.  We want to disable inbound pings instead.

sun (changes requested):

juanlopez4691:
Pushed a fixup with corrections and additional code. Can you review, please?

src/Admin.php
```diff
@@ -262,4 +262,10 @@ public static function admin_head() {
 EOD;
   }
 
+  /**
+   * Disables link notifications from other web sites.
+   */
+  public static function disableLinkNotifications() {
+    update_option('default_ping_status', 'closed');
```
sun: This should be implemented as a filter for the option name instead.

(Aside from that your proposed code would perform a write operation to the options table on every single request to an admin page – never do something like that.)

src/Plugin.php
```diff
@@ -65,12 +65,22 @@ public static function init() {
 
     add_shortcode('user-login-form', __NAMESPACE__ . '\UserLoginForm::getOutput');
 
+    // Disable self pingbacks.
+    add_action('pre_ping', __CLASS__ . '::pre_ping');
```
sun: What is a "self pingback"?

If this is about outbound pings, those can be disabled with the first option on `/wp-admin/options-discussion.php` as well as the last text field on `/wp-admin/options-writing.php`.

src/Plugin.php
```diff
@@ -65,12 +65,22 @@ public static function init() {
...
+
+    // Disable pingback functionality and remove x-pingback HTTP header
```
sun: Missing punctuation.

src/Plugin.php
```diff
@@ -179,4 +189,32 @@ public static function oembed_fetch_url($provider, $url, $args) {
...
+  }
+
+  /**
+   * @implements wp_headers
+   */
+  public static function wp_headers($headers) {
+    unset($headers['X-Pingback']);
+    return $headers;
+  }
 }
```
sun: Missing newline between method and end-of-class.

sun (changes requested):

src/Admin.php
juanlopez4691: 1. Ok, I will try the filter approach.
2. I am calling this function from the plugin activation hook. Is this not right then?

src/Plugin.php
juanlopez4691: The intention of this code is to avoid getting email notifications when linking some own content in an article. Perhaps it's out of the scope of the task, though. Shall I remove this then?

src/Admin.php
sun: Sorry about 2., I wrongly assumed a different invocation.

src/Plugin.php
sun: Yes, the goal (for now) is to just get rid of "pings from external sites end up as comments".

src/Admin.php
juanlopez4691: Accepting 2 as correct, I feel I should not set `default_ping_status` option with a filter, as it should be triggered in every frontend page rendering. Also, I find it somehow confusing having that option 'checked' in the site configuration but the code silently doing the opposite.

juanlopez4691:
Pushed some corrections. Review, please?

src/Admin.php
sun: The option should be forced to be disabled / not allow pingbacks.  The override should also affect existing sites.

The stored option value will not be visible in the backend, because the value will be always overridden.  If you want to prevent confusion, you can additionally disable/grey-out the option in the backend UI – which AFAIK should be easy to do, as all options are registered as admin settings and their form elements can be filtered individually.

sun (changes requested):
Would be great to finalize this soon with as little time as possible

juanlopez4691:
Pushed a new fixup to this PR. Can you review, please?

src/Plugin.php
sun: Do you plan to remove it?

src/Plugin.php
```diff
@@ -65,6 +65,28 @@ public static function init() {
...
+    add_filter('pre_option_default_pingback_flag', '__return_zero');
+
+    // Disables insertion of rel="pingback" tags.
+    add_filter('bloginfo_url', __CLASS__ . '::bloginfo_url', 10, 2);
+
+    // Unchecks 'allow trackbacks and pingbacks' in the Discussion meta box of post editing page.
+    add_action('post_comment_status_meta_box-options', __CLASS__ . '::post_comment_status_meta_box_options');
+
+    // Unchecks 'allow trackbacks and pingbacks' prior to save the post.
+    add_filter( 'wp_insert_post_data' , __CLASS__ . '::wp_insert_post_data' , '99', 2 );
```
sun: Bad coding style & wrong data type for 2nd argument.

src/Plugin.php
```diff
@@ -65,6 +65,28 @@ public static function init() {
...
+    add_filter('pre_update_default_ping_status', '__return_false');
+    add_filter('pre_option_default_ping_status', '__return_zero');
+    add_filter('pre_update_default_pingback_flag', '__return_false');
+    add_filter('pre_option_default_pingback_flag', '__return_zero');
+
+    // Disables insertion of rel="pingback" tags.
+    add_filter('bloginfo_url', __CLASS__ . '::bloginfo_url', 10, 2);
+
+    // Unchecks 'allow trackbacks and pingbacks' in the Discussion meta box of post editing page.
+    add_action('post_comment_status_meta_box-options', __CLASS__ . '::post_comment_status_meta_box_options');
```
sun: Isn't it sufficient to override the data value that is saved to the database (as below)?  Why do you need to additionally modify the metabox output?

src/Plugin.php
```diff
@@ -65,6 +65,28 @@ public static function init() {
...
+    add_filter('wp_headers', __CLASS__ . '::wp_headers');
+
+    // Disables admin discussion options for pingbacks.
+    add_filter('pre_update_default_ping_status', '__return_false');
+    add_filter('pre_option_default_ping_status', '__return_zero');
+    add_filter('pre_update_default_pingback_flag', '__return_false');
+    add_filter('pre_option_default_pingback_flag', '__return_zero');
+
+    // Disables insertion of rel="pingback" tags.
+    add_filter('bloginfo_url', __CLASS__ . '::bloginfo_url', 10, 2);
```
sun: This is too late. An empty value will cause an invalid HTTP header to be sent: https://github.com/WordPress/WordPress/blob/9fdbe6538e5718a801b2f4ae57260d2d1db4aae6/wp-includes/class-wp.php#L654

Instead you should filter the `pings_open` check: https://developer.wordpress.org/reference/functions/pings_open/

See also https://github.com/WordPress/WordPress/search?q=pings_open

src/Plugin.php
```diff
@@ -65,6 +65,28 @@ public static function init() {
...
 
+    // Disable self pingbacks.
+    add_action('pre_ping', __CLASS__ . '::pre_ping');
+
+    // Disable pingback functionality and remove x-pingback HTTP header.
+    add_filter('xmlrpc_methods', __CLASS__ . '::xmlrpc_methods');
+    add_filter('wp_headers', __CLASS__ . '::wp_headers');
+
+    // Disables admin discussion options for pingbacks.
+    add_filter('pre_update_default_ping_status', '__return_false');
```
sun: Normally it is sufficient to filter the option value; the update is not really relevant.

src/Plugin.php
```diff
@@ -65,6 +65,28 @@ public static function init() {
 
     add_shortcode('user-login-form', __NAMESPACE__ . '\UserLoginForm::getOutput');
 
+    // Disable self pingbacks.
+    add_action('pre_ping', __CLASS__ . '::pre_ping');
+
+    // Disable pingback functionality and remove x-pingback HTTP header.
+    add_filter('xmlrpc_methods', __CLASS__ . '::xmlrpc_methods');
+    add_filter('wp_headers', __CLASS__ . '::wp_headers');
```
sun: The header will not be added when using the `pings_open` filter as noted below.

sun (changes requested):

juanlopez4691:
Pushed a new fixup with the corrections you suggested.

src/Plugin.php
```diff
@@ -65,27 +65,15 @@ public static function init() {
 
     add_shortcode('user-login-form', __NAMESPACE__ . '\UserLoginForm::getOutput');
 
-    // Disable self pingbacks.
-    add_action('pre_ping', __CLASS__ . '::pre_ping');
-
-    // Disable pingback functionality and remove x-pingback HTTP header.
-    add_filter('xmlrpc_methods', __CLASS__ . '::xmlrpc_methods');
```
sun: Is the XMLRPC method removed through the `pings_open` filter, too?

src/Plugin.php
```diff
@@ -65,27 +65,15 @@ public static function init() {
...
     add_filter('pre_option_default_ping_status', '__return_zero');
-    add_filter('pre_update_default_pingback_flag', '__return_false');
     add_filter('pre_option_default_pingback_flag', '__return_zero');
 
     // Disables insertion of rel="pingback" tags.
-    add_filter('bloginfo_url', __CLASS__ . '::bloginfo_url', 10, 2);
-
-    // Unchecks 'allow trackbacks and pingbacks' in the Discussion meta box of post editing page.
-    add_action('post_comment_status_meta_box-options', __CLASS__ . '::post_comment_status_meta_box_options');
+    add_filter('pings_open', '__return_false', 20, 2);
```
sun: 1. I'm fairly sure `__return_false` takes only zero to one arguments.

1. The priority could be a bit higher, too; e.g. 100.

src/Plugin.php
```diff
@@ -65,27 +65,15 @@ public static function init() {
...
 
     // Unchecks 'allow trackbacks and pingbacks' prior to save the post.
-    add_filter( 'wp_insert_post_data' , __CLASS__ . '::wp_insert_post_data' , '99', 2 );
+    add_filter('wp_insert_post_data' , __CLASS__ . '::wp_insert_post_data', 99, 2);
```
sun: Your callback takes one argument only?

sun (changes requested):

src/Plugin.php
juanlopez4691: Ouch, you are right.

juanlopez4691:
@sun Just pushed the last corrections you suggested. Thanks for reviewing!

src/Plugin.php
```diff
@@ -69,11 +69,14 @@ public static function init() {
...
+    // Disable pingback functionality and remove x-pingback HTTP header.
+    add_filter('xmlrpc_methods', __CLASS__ . '::xmlrpc_methods');
+
     // Disables insertion of rel="pingback" tags.
-    add_filter('pings_open', '__return_false', 20, 2);
+    add_filter('pings_open', '__return_false', 100, 0);
 
     // Unchecks 'allow trackbacks and pingbacks' prior to save the post.
-    add_filter('wp_insert_post_data' , __CLASS__ . '::wp_insert_post_data', 99, 2);
+    add_filter('wp_insert_post_data' , __CLASS__ . '::wp_insert_post_data', 100, 2);
```
sun: You don't actually use the second argument, why do you specify it?

src/Plugin.php
```diff
@@ -190,9 +193,17 @@ public static function oembed_fetch_url($provider, $url, $args) {
   }
 
   /**
+   * @implements xmlrpc_methods
+   */
+  public static function xmlrpc_methods($methods) {
+    unset($methods['pingback.ping']);
```
sun: What about the second pingback method?
https://github.com/WordPress/WordPress/blob/339d8384755f0355c076fb7cf2e299cce79bd453/wp-includes/class-wp-xmlrpc-server.php#L149

src/Plugin.php
```diff
@@ -69,11 +69,14 @@ public static function init() {
     add_filter('pre_option_default_ping_status', '__return_zero');
     add_filter('pre_option_default_pingback_flag', '__return_zero');
 
+    // Disable pingback functionality and remove x-pingback HTTP header.
```
sun: 1. "X-Pingback" should be written in the lettercasing as it appears.

2. This comment appears to be a heading for all three filters that follow, not just the first one (the first filter is only about the HTTP header).  In closely connected cases like this, You can combine the other two comments into this first one as necessary (still retaining relevant details) and put all three filter registrations directly below this single comment.

sun (changes requested):
Almost done! :+1:

juanlopez4691:
Just pushed a fixup with the last corrections suggested. Thank you!

sun:
@juanlopez4691 Pushed some final adjustments, please review.

PR #13, opened 2017-11-08
Exclude subscribers from author dropdown list

juanlopez4691:


src/Admin.php
```diff
@@ -35,6 +35,9 @@ public static function init() {
     // Exposes SVG images in media library.
     add_filter('wp_prepare_attachment_for_js', __CLASS__ . '::wp_prepare_attachment_for_js');
     add_action('admin_head', __CLASS__ . '::admin_head');
+
+    // Excludes users with subscriber role from the author dropdown select list.
```
sun: You should include the reason _why_ you're doing this.

src/Admin.php
```diff
@@ -35,6 +35,9 @@ public static function init() {
...
+    add_filter('wp_dropdown_users_args', __CLASS__ . '::wp_dropdown_users_args', 10, 1);
```
sun: `1` is the default value for the number of callback arguments, you don't need to specify it.

src/Admin.php
```diff
@@ -262,4 +265,13 @@ public static function admin_head() {
 EOD;
   }
 
+  /**
+   * @implements wp_dropdown_users_args
+   */
+  public static function wp_dropdown_users_args($query_args) {
+    $query_args['who'] = 'author';
```
sun: Instead of setting the parameter, you need to use it as condition.  Do you understand why?

src/Admin.php
```diff
@@ -262,4 +265,13 @@ public static function admin_head() {
 EOD;
   }
 
+  /**
+   * @implements wp_dropdown_users_args
+   */
+  public static function wp_dropdown_users_args($query_args) {
```
sun: The argument can be type-hinted as an array.

sun (changes requested):

juanlopez4691:
Pushed a fixup with corrections. Review, please?

src/Admin.php
```diff
@@ -262,4 +266,15 @@ public static function admin_head() {
 EOD;
   }
 
+  /**
+   * @implements wp_dropdown_users_args
+   */
+  public static function wp_dropdown_users_args(array $query_args) {
+    if ($query_args['who'] !== 'authors') {
```
sun: 1. I do not understand how this currently proposed code is able to work; the condition needs to be negated if you want to match https://developer.wordpress.org/reference/functions/post_author_meta_box/

2. The parameter may not always be set, so the first condition should be `isset()`.

src/Admin.php
```diff
@@ -35,6 +35,10 @@ public static function init() {
     // Exposes SVG images in media library.
     add_filter('wp_prepare_attachment_for_js', __CLASS__ . '::wp_prepare_attachment_for_js');
     add_action('admin_head', __CLASS__ . '::admin_head');
+
+    // Excludes users with subscriber role from the author dropdown select list to avoid
+    // an extremely long list in the the case of sites with a large amount of subscribers.
```
sun: The large list is not really the problem – the problem is the page loading performance:
```diff
-    // Excludes users with subscriber role from the author dropdown select list to avoid
-    // an extremely long list in the the case of sites with a large amount of subscribers.
+    // Exclude subscribers from post author select options to prevent a performance
+    // slowdown on sites with large amounts of non-administrative registered users.
```

sun (changes requested):

juanlopez4691:
Fixed point 2 of your last comment. Don't get what you mean in point 1 of same comment, sorry. Businness directory plugin is setting `['who'] => 'all'` in a filter. I guess this is related to the functionality of subscribers being able to add new entries from the frontend.

src/Admin.php
sun: > Fixed point 2 of your last comment. Don't get what you mean in point 1 of same comment, sorry.
>
> The Businness directory plugin I'm using is setting ['who'] => 'all' in a filter. I guess this is related to the functionality of subscribers being able to add new entries from the frontend.
> 
> I checked the code in my local site and works OK.

core-standards is a generic plugin which needs to work on all WordPress sites.  Therefore, we have to care for the regular post author select options on a regular post/page first; i.e., on a screen like `/wp-admin/post.php?post=1234&action=edit` for post type `post` and `page`.

https://developer.wordpress.org/reference/functions/wp_dropdown_users/ clearly documents that `who` can only be empty or `authors`:

> 'who' (string) Which type of users to query. Accepts only an empty string or 'authors'.

…so that business directory plugin does not appear to pass correct parameters.

But as you mentioned, the plugin might actually try to announce that it's trying to offer all users as possible author options.  From a distance, that appears to make sense, as I assume the business directory entries will be authored by regular / non-administrative users?

juanlopez4691: >... as I assume the business directory entries will be authored by regular / non-administrative users?

Correct, this seems to be the intended functionality of the plugin. I assume then that we do not expect the site to work this way, right?

Perhaps I could narrow the condition in some way:
1. Detecting if the current page is business-directory-plugin related. Something like:
```
if ($post_type === 'wpbdp_listing' && $query_args['who'] !== 'authors') { ...
```
2. Checking for `$query_args['who'] !== 'authors' && $query_args['who'] !== ''`
3. Just checking for `$query_args['who'] === 'all'`

juanlopez4691: 2. Could be improved: `not_in_array($query_args['who'], ['authors', ''])`

sun: Unfortunately we can only care for the case of `$query_args['who'] === 'authors'` in core-standards, because that is the only case that can be safely adjusted on all WordPress sites.  But in case it helps in any way, core-standards could invoke a filter.

juanlopez4691: If I understand this correctly, and assuming `'authors'` and `''` are the ony "acceptable" values for `$query_args['who']`, then
```php
if (isset($query_args['who']) && not_in_array($query_args['who'], ['authors', '']) {
...
}
```
should be "safe" enough, right?

The other option would be implementing back the solution to this issue in the site custom plugin (bringing back the class BusinessDirectory I removed).

sun: Nope, it looks like you're stuck in a wrong interpretation.  Let me explain the parameter `who` once more to clear up that confusion:

The code that calls `wp_dropdown_users()` specifies the parameter `who` to inform all possibly running filters about the purpose of the select dropdown element.

If the dropdown is used for selecting the author of a (editorial) post, then the calling code should indicate that by specifying `who: authors`

If the dropdown does not have that limited and focused purpose, then the calling code should leave the parameter `who` empty, in order to indicate that all users will be possible and legit options in the dropdown.

The latter seems to be the case in your business directory.  Listings in there are created by regular users (in the front-end) and they become the author of their listings.  The dropdown for selecting a (different) author should therefore include regular users (subscribers) by default.  By specifying `who: all` it explicitly asks for all users, not just editorial post authors.

We cannot diverge from that expectation in core-standards.  The only change and optimization we could consider for core-standards would be to replace the entire dropdown with an autocomplete widget that uses an AJAX endpoint for searching matching users.

A quick search for "dropdown users autocomplete" yields e.g. https://wordpress.org/plugins/authors-autocomplete-meta-box/ (maintained on https://github.com/bamadesigner/authors-autocomplete-meta-box), which looks unnecessarily complex to me and seemingly invokes way too many filters… — if there is no readily available, solid and working solution for this yet, then we can consider to incorporate a solution into core-standards.

In the meantime, it would certainly make sense to move forward with this PR, but the condition needs to be changed into:
```diff
-    if (isset($query_args['who']) && $query_args['who'] !== 'authors') {
+    if (isset($query_args['who']) && $query_args['who'] === 'authors') {
```

sun (approved):
Thank you!

src/Admin.php
juanlopez4691: >The code that calls wp_dropdown_users() specifies the parameter who to inform all possibly running filters about the purpose of the select dropdown element.

I get that, sure. I think then we are trying to solve different things: I am thinking on the particular problem with the business-directory-plugin.

>The latter seems to be the case in your business directory. Listings in there are created by regular users (in the front-end) and they become the author of their listings. The dropdown for selecting a (different) author should therefore include regular users (subscribers) by default. By specifying who: all it explicitly asks for all users, not just editorial post authors.

Then the right approach would be setting the `wp_dropdown_users_args` filter in the site custom plugin in order to exclude subscribers but just in the backend, right?

PR #15, opened 2018-02-07
Added fast 404 error responses for non-existing uploads.

juanlopez4691:
To improve performance, a 404 error is returned instead of WordPress generating a file not found page.


src/Admin.php
```diff
@@ -39,6 +39,10 @@ public static function init() {
     // Exclude subscribers from post author select options to prevent a performance
     // slowdown on sites with large amounts of non-administrative registered users.
     add_filter('wp_dropdown_users_args', __CLASS__ . '::wp_dropdown_users_args');
+
+    // Copies .htaccess template file into uploads directory.
+    add_action('wp_upgrade', __CLASS__ . '::wp_upgrade', 10, 2);
+    add_action('wp_install', __CLASS__ . '::wp_install');
```
sun: 1. `wp_install` is fired when a new site is fully installed.  At that point, this plugin will not exist yet, so registering this filter is pointless.

1. You don't need to create and register a separate callback function if all arguments are ignored and all it does is to forward to another function.  You can register the final function instead.

1. The callback function for this special operation should be located in the Schema class.

1. These special actions should be registered from the main plugin file.

src/Admin.php
```diff
@@ -277,4 +281,26 @@ public static function wp_dropdown_users_args(array $query_args) {
...
+  public static function wp_install($wp_user) {
+    static::createUploadsHtaccessFile();
+  }
+
+  /**
+   * Copies .htaccess template file into uploads directory.
+   */
+  public static function createUploadsHtaccessFile() {
+    $upload_dir = explode('sites/', wp_upload_dir(NULL, FALSE)['basedir'])[0];
+    copy(dirname(__DIR__) . '/conf/.htaccess.uploads', $upload_dir . '/.htaccess');
```
sun: 1. Before overwriting an existing file, the code should check whether the existing file contains the content of the template file already.  Doesn't have to be identical, there may be additional lines above or below, but the template should be fully contained.  The method name should be adjusted accordingly from "create" into "ensure".

1. You also need to check whether the target file is writable.

1. A similar operation should be performed on the .gitignore file.

1. When operating on WP-CLI, there should be corresponding success or error messages, so that the user is aware that additional operations were (attempted to be) performed.

sun (changes requested):

juanlopez4691:
@sun Pushed a fixup. I hope I have covered all your suggestions. Code tested and worked (WP admin & WP-CLI).

sun (approved):
Cleaned, refactored, and simplified the implementation – limiting error reporting exclusively to WP-CLI for now.

This works correctly for me on new sites (not having the uploads .htaccess file yet) as well as existing sites (having one already).

There's one major caveat:  If the lines of the template were adjusted or modified in any way, then core-standards will duplicate them.  I can't see a simple way around that issue, but as it is a hypothetical one right now, I simply adjusted the positive confirmation message to include a _"Review this before committing!"_

juanlopez4691:
Thanks for your review!
> There's one major caveat: If the lines of the template were adjusted or modified in any way, then core-standards will duplicate them.

Yes, I saw this possible problem too, but I do not see an easy and robust solution to that either.

juanlopez4691:
@sun Now it is squash & merge time, right? Using the PR title as commit title?

sun (approved):

PR #16, opened 2018-05-10
Added Google Analytics tracking opt-out backend settings and frontend…

juanlopez4691:
… script injection.

Related to task [Core standards: Add Tracking opt-out for Google Analytics](https://app.asana.com/0/inbox/383242129844222/607229343598199/607239277828768).

- Added opt-out script backend admin menu entry and settings page.
- Added frontend opt-out script injection.

Estimated: 8h
Time left: 0.5h


src/OptOut.php
```diff
@@ -0,0 +1,146 @@
...
+   */
+  public static function admin_menu() {
+    add_submenu_page('options-general.php', __('WordPress Core Standards Settings', Plugin::L10N), __('WordPress Core Standards', Plugin::L10N), 'manage_options', 'core-standards-settings', [__CLASS__, 'renderSettings']);
+  }
+
+  /**
+   * Renders the settings page template.
+   */
+  public static function renderSettings() {
+    Plugin::renderTemplate(['templates/settings.php']);
```
sun: Can we add these settings to e.g. the "Reading" page instead, so that we can simply register the settings and do not need a custom HTML output / template at all?

src/OptOut.php
```diff
@@ -0,0 +1,146 @@
...
+class OptOut {
+
+  /**
+   * Default Google Analytics opt out script.
+   *
+   * @var string
+   */
+  private static $opt_out_script = <<<CODE
+<script>
+// Disable tracking if the opt-out cookie exists.
```
sun: This script should be in an asset JS file and the GA IDs should be read from JS settings variables of the global document scope.

In case you intentionally inlined the script for frontend loading performance reasons, then you can still do that by including the compiled script from the dist folder instead.

src/OptOut.php
```diff
@@ -0,0 +1,146 @@
...
+
+    add_action('admin_enqueue_scripts', __CLASS__ . '::admin_enqueue_scripts');
+  }
+
+  /**
+   * @implements init
+   */
+  public static function init() {
+    if (is_admin()) {
+      return;
```
sun: This check should be performed before registering the callback instead of here, so that the class/callback is not loaded in situations where it isn't needed.

src/OptOut.php
```diff
@@ -0,0 +1,146 @@
+<?php
+
+/**
+ * @file
+ * Contains \Netzstrategen\CoreStandards\OptOut.
+ */
+
+namespace Netzstrategen\CoreStandards;
+
+class OptOut {
```
sun: This appears to be all about Google Analytics, so why don't we call it like that?

Simply `GoogleAnalytics` or if you think that's too unspecific then `GoogleAnalyticsOptOut`

src/OptOut.php
```diff
@@ -0,0 +1,146 @@
...
+  public static function renderSettings() {
+    Plugin::renderTemplate(['templates/settings.php']);
+  }
+
+  /**
+   * @implements wp_enqueue_scripts.
+   */
+  public static function wp_enqueue_scripts() {
+    if (get_option('core_standards_optout_active')) {
+      wp_enqueue_script(Plugin::PREFIX . '-opt-out', Plugin::getBaseUrl() . '/assets/scripts/opt-out.js', ['jquery'], FALSE, TRUE);
```
sun: The compiled/optimized scripts from dist should be loaded instead of the source scripts from assets.

Also note that the compiled script files in dist are missing in this PR.

assets/scripts/admin-opt-out.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+(function ($) {
+  var $dest = $('#core-standards-optout-script');
+  // Restores the default opt-out script if textarea is empty.
+  if (!$dest.val().trim().length) {
+    load_default_script($dest);
```
sun: The Core Standards plugin is about sane & smart standards… — Why would I ever want to override the opt-out script with a custom one, editing in a textarea?  Wouldn't we improve the default script instead?

Was there a particular request or requirement for making it possible to override the default script?

assets/scripts/opt-out.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+(function ($) {
+  // Triggers out-out cookie creation when click on link.
+  $('.core-standards-optout-link').on('click', function(e) {
+    e.preventDefault();
+    gaOptout()
+    alert(core_standards_optout.confirmation_message);
```
sun: The part before the dot should be the namespace of the plugin; i.e., only `core_standards` and the `optout` part needs to move after the period.

assets/scripts/opt-out.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+(function ($) {
+  // Triggers out-out cookie creation when click on link.
+  $('.core-standards-optout-link').on('click', function(e) {
```
sun: The missing space after the keyword `function` should have triggered an eslint compile error – please double-check with @LucaPipolo whether the frontend build pipeline is correctly set up for the core-standards plugin.

languages/core-standards.pot
```diff
@@ -0,0 +1,126 @@
+#, fuzzy
+msgid ""
+msgstr ""
+"Plural-Forms: nplurals=INTEGER; plural=EXPRESSION;\n"
+"Project-Id-Version: Core Standards\n"
```
sun: `core-standards`

assets/scripts/admin-opt-out.js
juanlopez4691: @sun Although I don't really like the idea, this is defined as a "nice to have" requirement in the task summary. I  tried to implement that.

assets/scripts/opt-out.js
juanlopez4691: @sun As far as I see, there is no build system set for this plugin (no package.json, no gulpfile.json...).

src/OptOut.php
juanlopez4691: @sun There is no build pipeline set for core-standards plugin (I found this when I started implementing). If I shall set it up I will need to correct my initial time estimation.

src/OptOut.php
juanlopez4691: @sun I did it like this because I though it would allow us to add further settings to core-standards plugin in the future.

Of course I can move this settings to the "Reading" page, but I am not sure if this is not mixing apples & oranges. Considering this, can you confirm me you want me to do so? 😉

src/OptOut.php
juanlopez4691: > ... the GA IDs should be read from JS settings variables of the global document scope.

@sun Not sure if I get your point here. The GA IDs are being injected in the frontend as global vars. See https://github.com/netzstrategen/wordpress-core-standards/blob/fec68e4b9017ffeccef5cdbb3caa3e4c147c2613/src/OptOut.php#L72-L90

juanlopez4691: > In case you intentionally inlined the script for frontend loading performance reasons, then you can still do that by including the compiled script from the dist folder instead.

Not for performance reasons, but to be sure the script is loaded before any tracking script and to implement that "nice to have feature" consisting of "Client should be able to maintain the script, which handles the opt-out from the backend" (from task summary).

src/OptOut.php
juanlopez4691: @sun Changed it to `TrackingOptOut` instead of `GoogleAnalyticsOptOut` to left it open to cover other tracking systems in the future (see "nice to have features" in the task summary).

juanlopez4691:
@sun Added a yarn + gulp build pipeline.

src/OptOut.php
sun: We generally try to avoid adding new option pages, as our plugins are usually very small.

But especially in this case it makes sense to add to existing pages, as the core-standards plugin only corrects/enhances the defaults of WordPress Core (and in an ideal world, Core would ship with everything we do in core-standards already) and doesn't really provide a very unique thing of its own.

assets/scripts/admin-opt-out.js
sun: I do not like the idea either and would recommend to remove it.  The only configuration I expect from this feature are the GA IDs.

languages/core-standards-de_DE_formal.po
```diff
@@ -0,0 +1,126 @@
+msgid ""
+msgstr ""
+"Project-Id-Version: Core Standards\n"
```
sun: Also needs to be changed here

core-standards.php
```diff
@@ -38,5 +38,12 @@ function classloader($class) {
...
 
 add_action('widgets_init', __NAMESPACE__ . '\Widgets\UserLoginFormWidget::init');
+add_action('plugins_loaded', __NAMESPACE__ . '\Plugin::loadTextdomain');
 add_action('init', __NAMESPACE__ . '\Plugin::init', 20);
 add_action('admin_init', __NAMESPACE__ . '\Admin::init');
+
+add_action('admin_menu', __NAMESPACE__ . '\TrackingOptOut::admin_menu');
+add_action('admin_init', __NAMESPACE__ . '\TrackingOptOut::admin_init');
+if (!is_admin()) {
+  add_action('init', __NAMESPACE__ . '\TrackingOptOut::init');
```
sun: Normally, the main plugin file only has the minimal hooks for distributing into the main controllers Schema, Admin, Plugin, and only those controllers may distribute further.

In this case it might make sense to do it differently though, as the callbacks need to run in more special priorities.

package.json
```diff
@@ -0,0 +1,34 @@
...
+  },
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/netzstrategen/wordpress-core-standards.git"
+  },
+  "author": "netzstrategen",
+  "license": "ISC",
```
sun: The license of our public plugins should be GPLv2 or later, as in the main plugin file.

assets/scripts/admin-opt-out.js
juanlopez4691: @sun Added a comment to the task recommending Peter about discarding this feature.

src/OptOut.php
sun: Note: Unless I'm mistaken, when adding to existing pages, the entire template for the administrative configuration page is obsolete.  The form elements are specified along with the call to register_setting() already.

Tried to find existing work but only found this (which isn't exactly what I mean here): https://github.com/netzstrategen/wordpress-dfp/blob/master/src/Admin.php#L43-L50

juanlopez4691:
@sun Applied suggested corrections and some refactoring (removed the "edit opt-out script" feature).

src/TrackingOptOut.php
```diff
@@ -74,15 +98,14 @@ public static function wp_head() {
     }
     // Converts any non-word character into a comma and removes white-space characters
     // (except '-') from the string to get a comma-separated list of Google Analytics IDs.
-    $optout_ids = preg_replace(['@\W^-|\n@', '@\s@'], [',', ''], esc_attr(get_option('core_standards_optout_ids')));
+    $optout_ids = preg_replace(['@\W*^-|\s+|:|;@'], [','], esc_attr(get_option('core_standards_optout_ids')));
```
sun: This appears to be a unnecessarily complex way of splitting a string by whitespace:

```php
$optout_ids = preg_split('/[\s,]+/', get_option('core_standards_optout_ids'));
```
This is also documented as most common example on http://php.net/manual/en/function.preg-split.php

This generates an array already, so you do not need the additional `explode()` below either.

`esc_attr()` also appears to be unnecessary, as the value are not output as part of HTML; they are serialized into JSON instead.

src/TrackingOptOut.php
```diff
@@ -101,30 +124,23 @@ public static function registerOptOutShortcode ($atts, $content = NULL) {
...
+
   /**
    * @implements wp_enqueue_scripts.
    */
   public static function wp_enqueue_scripts() {
-    if (get_option('core_standards_optout_active')) {
-      wp_enqueue_script(Plugin::PREFIX . '-opt-out', Plugin::getBaseUrl() . '/dist/scripts/opt-out.js', ['jquery'], FALSE, TRUE);
-      wp_localize_script(Plugin::PREFIX . '-opt-out', 'core_standards', ['optout_confirmation_message' => __('Google Analytics tracking has been disabled.', Plugin::L10N)]);
-    }
+    wp_enqueue_script(Plugin::PREFIX . '-opt-out', Plugin::getBaseUrl() . '/dist/scripts/opt-out.js', ['jquery'], FALSE, TRUE);
```
sun: 1. `enqueueScriptsFirstHead()` above uses `-ga-opt-out` instead of `-opt-out` – typo or intended?  If it's intended, then the difference between both is not clear at all and they should have clearer names.

2. Some code uses `optout` (varnames, option names, etc) instead of `opt-out` (both in identifier and filename).  That's not good for discovery / search purposes.  Can you make that consistent?  (Note that the class name `...OptOut` is converted into `...opt_out` with `uncamelize()`)

src/TrackingOptOut.php
```diff
@@ -9,60 +9,84 @@
...
+    add_settings_field(
+      'core_standards_optout_ids',
+      __('Google Analytics IDs', Plugin::L10N),
+      __CLASS__ . '::optOutGoogleIdsFieldRender',
+      'general',
+      Plugin::PREFIX . '-optout'
+    );
+
+    add_settings_field(
+      'core_standards_optout_active',
```
sun: This separate activation setting looks unnecessary.  The feature should automatically be toggled depending on whether GA IDs were configured — there is no point in enabling the feature without GA IDs.

src/TrackingOptOut.php
```diff
@@ -9,60 +9,84 @@
...
+    }
+  }
+
+  /**
+   * Adds Opt-Out section settings fields.
+   */
+  public static function addSettingsFields() {
+    add_settings_section(
+      Plugin::PREFIX . '-optout',
+      __('WordPress Core Standards Settings', Plugin::L10N),
```
sun: 1. A separate section for a single setting is a bit overkill.  New sections should only be added wisely.

1. If/when adding new sections, they should follow the UI guidelines of WordPress; i.e., their labels/titles should focus on functionality/perspective instead of a technical implementation detail (a plugin name).

1. The Reading settings page still appears to be a much better fit than the general settings page: https://gartenmoebelcompany.de/wp-admin/options-reading.php – for example, below the "Search engine visibility" setting.

src/TrackingOptOut.php
juanlopez4691: @sun I want to keep the separate activation control for two reasons:
1. I don't like having a single interface element (GA IDs text box) doing two different functions; that's not good UX, IMO.
2. This way opt-out can be disabled without having to wipe the GA IDs list. No need to remember/searching for the GA IDs if user deactivated opt-aut and wants to reactivate it later on.

What I will do is adding a condition to disable opt-out if the GA IDs is empty.

juanlopez4691:
@sun Implemented your last suggestions. Thanks for your review.

juanlopez4691:
@sun Implemented your last suggestions. Thanks for your review.

languages/core-standards-de_DE_formal.po
```diff
@@ -0,0 +1,126 @@
...
+"X-Poedit-WPHeader: core-standards.php\n"
+"X-Poedit-SourceCharset: UTF-8\n"
+"X-Poedit-KeywordsList: __;_e;_n:1,2;_x:1,2c;_ex:1,2c;_nx:4c,1,2;esc_attr__;"
+"esc_attr_e;esc_attr_x:1,2c;esc_html__;esc_html_e;esc_html_x:1,2c;"
+"_n_noop:1,2;_nx_noop:3c,1,2;__ngettext_noop:1,2\n"
+"X-Poedit-SearchPath-0: .\n"
+"X-Poedit-SearchPathExcluded-0: *.js\n"
+
+#: src/OptOut.php:114 templates/settings.php:10
+msgid "WordPress Core Standards Settings"
```
sun, line 25: Outdated gettext files, these strings no longer exist in the code.

assets/scripts/enable-opt-out.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+(function ($) {
+  // Triggers out-out cookie creation when click on link.
+  $('.core-standards-opt_out-link').on('click', function (e) {
```
sun: Minor: The underscore is an alien in the class name; what made you choose it?  `opt_out` should be `opt-out` for consistency.

package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "core-standards",
+  "version": "1.0.0",
```
sun: The version number in package.json should be identical to the version in the main plugin PHP file.

src/TrackingOptOut.php
```diff
@@ -0,0 +1,141 @@
...
+      'default'
+    );
+
+    register_setting(
+      'reading',
+      'core_standards_opt_out_active'
+    );
+    add_settings_field(
+      'core_standards_opt_out_active',
+      __('Activate Opt-Out', Plugin::L10N),
```
sun: 1. I still don't see why we need two settings to toggle one thing… especially because this is about standards; i.e., we assume that all sites are going to need it, so there's little point in making it optional as our initial assumption is that every site needs it. — It's your call… just pointing out once more that the math doesn't add up for me.

1. If you keep two options instead of one, then the first option must be the checkbox to toggle the whole thing on and off. Children always come after their parents. That's UX 101.

src/TrackingOptOut.php
```diff
@@ -0,0 +1,141 @@
...
+  public static function optOutGoogleIdsFieldRender() {
+    echo '<textarea id="core-standards-opt_out-ids" name="core_standards_opt_out_ids" rows="8" cols="40">' .
+      esc_attr(get_option('core_standards_opt_out_ids')) . '</textarea>';
+  }
+
+  /**
+   * Renders Google Analytics Opt-Out activation checkbox field.
+   */
+  public static function optOutActiveFieldRender() {
+    // Deactivates opt-out if Google Analytics IDs fiels is empty.
```
sun: Typo in "fiels"

src/TrackingOptOut.php
```diff
@@ -0,0 +1,141 @@
...
+    if (!get_option('core_standards_opt_out_ids')) {
+      update_option('core_standards_opt_out_active', FALSE);
```
sun: Performing a write operation during a read request is an absolute no-go.  Never ever do this.

If there is a dependency between two options, then you need to adjust the POST values before they are processed by the Settings API.

(Of course, it would be much easier to simply have just one value, but…)

sun (changes requested):
Let's get this done today! 👍

juanlopez4691:
@sun Pushed a fixup with your suggested corrections. Thanks a lot for your review.

sun (approved):
Thank you!

PR #17, opened 2018-05-23
Added youtube nocookie oembed conversion for GDPR data privacy.

sun:
Based on
- https://github.com/nowherecoding/youtube-nocookie/blob/master/youtube-nocookie.php
- https://gdgts.de/wordpress-youtube-embed/

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -43,7 +43,7 @@ class Plugin {
     // Ensure that Facebook embeds do not exceed available content width.
     add_filter('oembed_fetch_url', __CLASS__ . '::oembed_fetch_url', 10, 3);
 
-    // Add wrapper to oEmbed elements.
+    // Enforce youtube-nocookie and add wrapper to oEmbed elements.
     add_filter('embed_oembed_html', __CLASS__ . '::embed_oembed_html', 10, 4);
 
     // Allow SVG files in media library.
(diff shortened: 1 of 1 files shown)


src/Plugin.php
```diff
@@ -93,9 +93,13 @@ public static function oembed_providers(array $providers) {
   }
 
   /**
-   * Wrap oEmbed output for styling purposes.
+   * Convert YouTube embeds and wrap oEmbed output for styling purposes.
```
juanlopez4691, line 96: Shouldn't it be "Converts" and "wraps"?

src/Plugin.php
```diff
@@ -43,7 +43,7 @@ public static function init() {
     // Ensure that Facebook embeds do not exceed available content width.
     add_filter('oembed_fetch_url', __CLASS__ . '::oembed_fetch_url', 10, 3);
 
-    // Add wrapper to oEmbed elements.
+    // Enforce youtube-nocookie and add wrapper to oEmbed elements.
```
juanlopez4691, line 46: Shouldn't it be "Adds"?

juanlopez4691 (approved):

src/Plugin.php
sun: In the other case yes, but not here. The verb in third person form only applies to PHPDoc summary lines, not to inline comments.

A function does something for you (because you invoke it).

Regular code in a function body is doing something for some reason that is being documented in inline comments.

juanlopez4691: Thanks @sun! I will try to remember that.

ChristophKnopf:
Hi, what is the status here? This ticket is time critical.

PR #18, opened 2018-10-16
Added domain for Google Analytics opt-out cookie.

tmoehrmann:
@cknopf i will test if the cookie will be properly set, again, when this is on live.

### DIFF
--- a/assets/scripts/google-analytics-opt-out.js
+++ b/assets/scripts/google-analytics-opt-out.js
@@ -14,7 +14,7 @@ function gaOptout() {
   var disableStr;
   for (var i = 0; i < window.gaIdList.length; i++) {
     disableStr = 'ga-disable-' + window.gaIdList[i];
-    document.cookie = disableStr + '=true; expires=Thu, 31 Dec 2099 23:59:59 UTC; path=/; domain=.' + core_standards_opt_out_cookie_domain;
+    document.cookie = disableStr + '=true; expires=Thu, 31 Dec 2099 23:59:59 UTC; path=/; domain=.' + core_standards_opt_out.cookie_domain;
     window[disableStr] = true;
   }
 }
--- a/assets/scripts/google-analytics-opt-out.js
+++ b/assets/scripts/google-analytics-opt-out.js
@@ -12,10 +12,9 @@ for (var i = 0; i < window.gaIdList.length; i++) {
  */
 function gaOptout() {
   var disableStr;
-  var domainName = cookieDomain.domainValue;
   for (var i = 0; i < window.gaIdList.length; i++) {
     disableStr = 'ga-disable-' + window.gaIdList[i];
-    document.cookie = disableStr + '=true; expires=Thu, 31 Dec 2099 23:59:59 UTC; path=/; domain=.' + domainName;
+    document.cookie = disableStr + '=true; expires=Thu, 31 Dec 2099 23:59:59 UTC; path=/; domain=.' + core_standards_opt_out_cookie_domain;
     window[disableStr] = true;
   }
 }
--- a/assets/scripts/google-analytics-opt-out.js
+++ b/assets/scripts/google-analytics-opt-out.js
@@ -12,9 +12,10 @@ for (var i = 0; i < window.gaIdList.length; i++) {
  */
 function gaOptout() {
   var disableStr;
+  var domainName = cookieDomain.domainValue;
   for (var i = 0; i < window.gaIdList.length; i++) {
     disableStr = 'ga-disable-' + window.gaIdList[i];
-    document.cookie = disableStr + '=true; expires=Thu, 31 Dec 2099 23:59:59 UTC; path=/';
+    document.cookie = disableStr + '=true; expires=Thu, 31 Dec 2099 23:59:59 UTC; path=/; domain=.' + domainName;
     window[disableStr] = true;
   }
 }
--- a/dist/scripts/google-analytics-opt-out.js
+++ b/dist/scripts/google-analytics-opt-out.js
@@ -1,8 +1,9 @@
 "use strict";
 
 function gaOptout() {
-    for (var i, t = 0; t < window.gaIdList.length; t++) i = "ga-disable-" + window.gaIdList[t], 
-    document.cookie = i + "=true; expires=Thu, 31 Dec 2099 23:59:59 UTC; path=/", window[i] = !0;
+    for (var i, o = cookieDomain.domainValue, a = 0; a < window.gaIdList.length; a++) i = "ga-disable-" + window.gaIdList[a], 
+    document.cookie = i + "=true; expires=Thu, 31 Dec 2099 23:59:59 UTC; path=/; domain=." + o, 
+    window[i] = !0;
 }
 
 for (var i = 0; i < window.gaIdList.length; i++) {
--- a/dist/scripts/google-analytics-opt-out.js
+++ b/dist/scripts/google-analytics-opt-out.js
@@ -2,7 +2,7 @@
 
 function gaOptout() {
     for (var i, o = 0; o < window.gaIdList.length; o++) i = "ga-disable-" + window.gaIdList[o], 
-    document.cookie = i + "=true; expires=Thu, 31 Dec 2099 23:59:59 UTC; path=/; domain=." + core_standards_opt_out_cookie_domain, 
+    document.cookie = i + "=true; expires=Thu, 31 Dec 2099 23:59:59 UTC; path=/; domain=." + core_standards_opt_out.cookie_domain, 
     window[i] = !0;
 }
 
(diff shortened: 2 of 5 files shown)


tmoehrmann:
I had to add the domain, so this cookie will be set across our subdomains too, just in case you wondered why i did this.

tmoehrmann:
fixup: removed trailing dot from domain to fit newer RFC 6265 Standard. First commit has not been viable for bare domain aka bnn.de.

assets/scripts/google-analytics-opt-out.js
```diff
@@ -14,7 +14,7 @@ function gaOptout() {
   var disableStr;
   for (var i = 0; i < window.gaIdList.length; i++) {
     disableStr = 'ga-disable-' + window.gaIdList[i];
-    document.cookie = disableStr + '=true; expires=Thu, 31 Dec 2099 23:59:59 UTC; path=/';
+    document.cookie = disableStr + '=true; domain=bnn.de; expires=Thu, 31 Dec 2099 23:59:59 UTC; path=/';
```
sun: The custom domain cannot be hardcoded. The backend would need to pass the value to use for the cookie domain as a JS setting.

sun (changes requested):

assets/scripts/google-analytics-opt-out.js
```diff
@@ -12,9 +12,15 @@ for (var i = 0; i < window.gaIdList.length; i++) {
  */
 function gaOptout() {
   var disableStr;
+  var domainName;
+  if (location.host.indexOf('www.') === 0) {
```
sun: As mentioned previously, the code should use the cookie domain of the backend, because this can be configured explicitly to be e.g. the second level domain (e.g., example.com) when the site runs on a third or fourth level domain (e.g., shop.example.com or my.shop.example.com), in order to make the cookies (primarily the user session) available to other applications.

See https://codex.wordpress.org/Editing_wp-config.php#Set_Cookie_Domain

If you do not know how to pass a value from the backend to the frontend as a JavaScript setting, then we can move forward here by using the value of `document.domain` as an interim solution, even though that value is not really correct in all cases.

Implementing the final fix (the correct cookie domain from the backend) would be preferred.

sun (changes requested):

tmoehrmann:
@sun im not sure, why my addition doesnt get compiled properly. Thx for giving me some feedback on the changes i did.

tmoehrmann:
@sun Hi sun, i used the cookie_domain now, im not sure if the location (TrackingOptout.php) for enqueue-stuff is proper but i think its ok to add this to thi file. As for the COOKIE_DOMAIN, will we have to use .bnn.de or bnn.de?

sun:
As per [RFC 2109](https://tools.ietf.org/html/rfc2109), to have a cookie available to all subdomains, you must put a `.` in front of the domain.

assets/scripts/google-analytics-opt-out.js
```diff
@@ -12,9 +12,10 @@ for (var i = 0; i < window.gaIdList.length; i++) {
  */
 function gaOptout() {
   var disableStr;
+  var domainName = cookieDomain.domainValue;
```
sun: No need for this copy into a variable, you can directly access the original value in the loop.

dist/scripts/google-analytics-opt-out.js
```diff
@@ -1,8 +1,9 @@
 "use strict";
 
 function gaOptout() {
-    for (var i, t = 0; t < window.gaIdList.length; t++) i = "ga-disable-" + window.gaIdList[t], 
-    document.cookie = i + "=true; expires=Thu, 31 Dec 2099 23:59:59 UTC; path=/", window[i] = !0;
+    for (var i, o = cookieDomain.domainValue, a = 0; a < window.gaIdList.length; a++) i = "ga-disable-" + window.gaIdList[a],
```
sun: This appears to be an outdated compiled file.

src/TrackingOptOut.php
```diff
@@ -115,6 +115,10 @@ public static function enqueueScriptsFirstHead() {
   public static function wp_enqueue_scripts() {
     wp_enqueue_script(Plugin::PREFIX . '-enable-opt_out', Plugin::getBaseUrl() . '/dist/scripts/enable-opt-out.js', ['jquery'], FALSE, TRUE);
     wp_localize_script(Plugin::PREFIX . '-enable-opt_out', 'core_standards', ['opt_out_confirmation_message' => __('Google Analytics tracking has been disabled.', Plugin::L10N)]);
+
+    wp_enqueue_script('ga_optout', plugins_url() . '/core-standards/dist/scripts/google-analytics-opt-out.js');
```
sun: To my knowledge the script is enqueued elsewhere already?

src/TrackingOptOut.php
```diff
@@ -115,6 +115,10 @@ public static function enqueueScriptsFirstHead() {
   public static function wp_enqueue_scripts() {
     wp_enqueue_script(Plugin::PREFIX . '-enable-opt_out', Plugin::getBaseUrl() . '/dist/scripts/enable-opt-out.js', ['jquery'], FALSE, TRUE);
     wp_localize_script(Plugin::PREFIX . '-enable-opt_out', 'core_standards', ['opt_out_confirmation_message' => __('Google Analytics tracking has been disabled.', Plugin::L10N)]);
+
+    wp_enqueue_script('ga_optout', plugins_url() . '/core-standards/dist/scripts/google-analytics-opt-out.js'); 
+    $cookiedomain = array('domainValue' => COOKIE_DOMAIN);
+    wp_localize_script('ga_optout', 'cookieDomain', $cookiedomain);
```
sun: 1. The first argument registers the script with a custom identifier within WordPress.  It needs to start with the plugin's namespace.  See line 116/117 above.

2. The second argument is the name of the variable that will be exposed to the window/DOM document.  It needs to be the JavaScript-compatible variation of the plugin's namespace.  See line 117 above.

3. A better property name would be `cookieDomain` instead of `domainValue`.  The separate `$cookiedomain` variable does not seem to be necessary, as the array can be inlined.  And the array declaration should use the new short syntax.

4. The final cookie domain value including leading dot should be prepared here already: `['cookieDomain' => '.' . ltrim(COOKIE_DOMAIN, '.')]`

sun (changes requested):

src/TrackingOptOut.php
```diff
@@ -107,14 +107,15 @@ public static function renderSettings() {
...
+    wp_localize_script(Plugin::PREFIX . '-google_analytics_opt_out', 'core_standards_opt_out_cookie_domain', '.' . ltrim(COOKIE_DOMAIN, '.'));
   }
 
   /**
    * @implements wp_enqueue_scripts.
    */
   public static function wp_enqueue_scripts() {
     wp_enqueue_script(Plugin::PREFIX . '-enable-opt_out', Plugin::getBaseUrl() . '/dist/scripts/enable-opt-out.js', ['jquery'], FALSE, TRUE);
-    wp_localize_script(Plugin::PREFIX . '-enable-opt_out', 'core_standards', ['opt_out_confirmation_message' => __('Google Analytics tracking has been disabled.', Plugin::L10N)]);
+    wp_localize_script(Plugin::PREFIX . '-enable-opt_out', 'core_standards_opt_out_confirmation_message', 'Google Analytics tracking has been disabled.', Plugin::L10N);
```
sun:
```suggestion
    wp_localize_script(Plugin::PREFIX . '-enable-opt_out', 'core_standards_opt_out_confirmation_message', __('Google Analytics tracking has been disabled.', Plugin::L10N));
```

We lost the string translation here.

sun (changes requested):
👍 Thanks, only one tiny issue left!

assets/scripts/google-analytics-opt-out.js
```diff
@@ -14,7 +14,7 @@ function gaOptout() {
   var disableStr;
   for (var i = 0; i < window.gaIdList.length; i++) {
     disableStr = 'ga-disable-' + window.gaIdList[i];
-    document.cookie = disableStr + '=true; expires=Thu, 31 Dec 2099 23:59:59 UTC; path=/';
+    document.cookie = disableStr + '=true; expires=Thu, 31 Dec 2099 23:59:59 UTC; path=/; domain=.' + core_standards_opt_out.cookie_domain;
```
ChristophKnopf, line 17: This should generate `.bnn.de` but does `subdomain + .bnn.de` and should also work for jobs.bnn.de where no cookie ist set.

ChristophKnopf (changes requested):

ChristophKnopf:
Maybe we have to discuss how cookies should be handled here because this don't have to be the way you want to treat cookies in your plugin. Maybe to think about a little configuration to be more flexible?

rodrigoaguilera:
Since I do not have access to the PR's branch I resumed the work here
https://github.com/netzstrategen/wordpress-core-standards/pull/24

sun:
Duplicate of #24

PR #20, opened 2018-11-13
Added logo url filter for websites that can not use login CSS file.

LucaPipolo:


colourgarden (approved):
Not tested myself but we reviewed pretty thoroughly together already 👍

PR #22, opened 2018-12-04
Added cookie notice.

fabianmarz:
This PR adds a footer cookie notice displayed on all pages. It uses local storage to check wether the notice should be shown. This will prevent missing cache hits in our (and probably other) Varnish configurations.

Unclear:
- We currently only have `.pot` plugin files, but no `.mo` files, why?
- Shall we make the code more hookable, or split it more and include one file per style, markup, script so the markup can be overridden in particular without necessarily overriding the script/styles, too?


sun:
> 1. We currently only have .pot plugin files, but no .mo files, why?

Not sure what you mean, there all files seem to exist here: https://github.com/netzstrategen/wordpress-core-standards/tree/master/languages

> 2. Shall we make the code more hookable, or split it more and include one file per style, markup, script so the markup can be overridden in particular without necessarily overriding the script/styles, too?

All of the code was intentionally inlined into the HTML to prevent a short moment of flickering in case the DOM and CSS is loaded and rendered before the JS is parsed and executed.

What we could do is to move the CSS/JS code into individual files in the assets folder – so that they can be compiled (and so we get syntax highlighting and so on) – but then `include` their content back into the inlined output.  However, we would mainly gain the language editing benefits right now, as the compiled code needs to be committed and we therefore cannot enable optimizations for the compilers, so the compiled code would look almost identical to the source code.  Thus, nice-to-have for syntax highlighting, but no other gains.

3. Note that we should test this both on the originating project site (and disable the previous implementation) as well as on another/vanilla site to confirm that it works correctly in all cases. (Can you link to the two test sites in Asana?)

fabianmarz:
1. Was wondering as there are no non-formal files 🤔 
2. Yes, first had the style in a separate file but thought it might be bit overkill, so I wanted to get an opinion about this before. IMO we can postpone this.
3. The implementation done here strips some CSS code which should be implemented from the project's CSS file itself. One enhancement could be to add a hook to inject some more "inline" CSS code to not rely on a requested file.

sun:
> 3\. One enhancement could be to add a hook to inject some more "inline" CSS code to not rely on a requested file.

👍 We'll already need that for re-establishing the current styling on the originating site, right?

**EDIT:** I'd suggest an action hook to add more styles (as overrides) to keep it simple for now.

fabianmarz:
Currently I just resolved this by adding the styles to the general style file. Will add a hook instead.

templates/cookie-notice.php
```diff
@@ -0,0 +1,42 @@
...
+    left: 0;
+    right: 0;
+    z-index: 100000;
+    padding: 0.5em;
+    text-align: center;
+    background-color: #000;
+    color: #fff;
+  }
+
+  <?= do_action('core_standards/cookie_notice/css') ?>
```
sun: The hook name `css` is the only sensible name for a filter (in case we add a filter in the future), so I would use a more action-y name like `after_css` or `css_extend`:
```suggestion
  <?= do_action('core_standards/cookie_notice/css_extend') ?>
```

templates/cookie-notice.php
```diff
@@ -0,0 +1,42 @@
...
+</style>
+
+<div id="cookie-notice" role="banner">
+  <span><?= __('Diese Website verwendet Cookies. Wenn Sie sich weiterhin auf dieser Seite aufhalten, gehen wir von Ihrem Einverständnis aus.', Plugin::L10N) ?></span>
```
sun:
```suggestion
  <span><?= __('This website uses cookies to give you the best possible service. By continuing to use the website, you agree to the use of cookies.', Plugin::L10N) ?></span>
```

templates/cookie-notice.php
```diff
@@ -0,0 +1,42 @@
...
+  <a href="#" id="cookies-accept" class="btn"><?= __('Ok') ?></a>
```
sun:
```suggestion
  <a href="#" id="cookies-accept" class="btn"><?= __('Okay') ?></a>
```

templates/cookie-notice.php
```diff
@@ -0,0 +1,42 @@
...
+  <a href="<?= apply_filters('core_standards/cookie_notice/policy_link', '/datenschutzerklaerung') ?>" target="_self" class="btn"><?= __('Mehr Infos', Plugin::L10N) ?></a>
```
sun:
```suggestion
  <a href="<?= apply_filters('core_standards/cookie_notice/policy_link', '/privacy') ?>" target="_self" class="btn"><?= __('Learn more', Plugin::L10N) ?></a>
```

src/Plugin.php
```diff
@@ -79,6 +79,9 @@ public static function init() {
     // Add teaser image to RSS feeds.
     add_action('rss2_item', __NAMESPACE__ . '\Feed::rss2_item');
 
+    // Output client-side cookie notice on all pages.
+    add_action('wp_footer', __CLASS__ . '::wp_footer');
```
```suggestion
    // Use a slightly higher weight to prevent the bar being output
    // before other footer content.
    add_action('wp_footer', __CLASS__ . '::wp_footer', 12);
```

templates/cookie-notice.php
fabianmarz: First accepted this but reverted it as 1. it's fine to go with `OK` and 2. it's used in other language files in WP, too.

sun (approved):

PR #23, opened 2018-12-17
Added a method to ensure indices exist on tables.

rodrigoaguilera:
Added autoload index for the options table.

https://app.asana.com/0/765126798656082/929020052075702

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -27,11 +27,11 @@ class Plugin {
   const L10N = self::PREFIX;
 
   /**
-   * Cron event name for checking database indices.
+   * Cron event name for checking database indexes.
    *
    * @var string
    */
-  const CRON_EVENT_ENSURE_INDICES = Plugin::PREFIX . '/cron/ensure-indices';
+  const CRON_EVENT_ENSURE_INDEXES = Plugin::PREFIX . '/cron/ensure-indexes';
 
 
   /**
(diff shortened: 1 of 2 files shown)


src/Plugin.php
```diff
@@ -26,6 +26,14 @@ class Plugin {
    */
   const L10N = self::PREFIX;
 
+  /**
+   * Cron event name for checking database indices.
+   *
+   * @var string
+   */
+  const CRON_EVENT_ENSURE_INDICES = Plugin::PREFIX . '/cron/ensure-indices';
```
```suggestion
  const CRON_EVENT_ENSURE_INDEXES = Plugin::PREFIX . '/cron/ensure-indexes';
```
Both indices and indexes are valid plural forms, but I'd recommend to use indexes, because developers will probably try to search the code for "index".

src/Plugin.php
```diff
@@ -264,4 +277,30 @@ public static function renderTemplate(array $template_subpathnames, array $varia
     }
   }
 
+  /**
+   * Cron event callback to ensure proper database indices.
+   */
+  public static function cron_ensure_indices() {
```
sun: As there will be a few more indexes, I'd recommend to move this code here (without the triggering code in init) either into a new Database class or the existing Schema class.

src/Plugin.php
```diff
@@ -264,4 +277,30 @@ public static function renderTemplate(array $template_subpathnames, array $varia
...
+    self::ensureIndex('options', 'autoload', '(autoload)');
```
```suggestion
    self::ensureIndex('options', 'autoload', 'autoload');
```
For simplicity and readability, I'd suggest to remove the outer parentheses, as they are always required around the definition, and the possible final index options and copying/locking options are seldomly/never used:
https://dev.mysql.com/doc/refman/8.0/en/create-index.html#create-index-options

sun (approved):
👍 Looks good to me and ready to go if it works!

Perhaps test and confirm with @ChristophKnopf & @florent-zambon

rodrigoaguilera:
For merging it to bnn it will need another PR so let's merge it into the plugin first

PR #24, opened 2019-01-11
Added custom cookie domain support for Google Analytics opt-out.

rodrigoaguilera:
https://app.asana.com/0/74314752394866/724856314072534

### DIFF
--- a/assets/scripts/enable-opt-out.js
+++ b/assets/scripts/enable-opt-out.js
@@ -5,6 +5,6 @@
   $('.core-standards-opt-out-link').on('click', function (e) {
     e.preventDefault();
     gaOptout();
-    alert(core_standards.opt_out_confirmation_message);
+    alert(core_standards_opt_out_toggle.confirmation_message);
   });
 })(jQuery);
--- a/assets/scripts/google-analytics-opt-out.js
+++ b/assets/scripts/google-analytics-opt-out.js
@@ -14,7 +14,7 @@ function gaOptout() {
   var disableStr;
   for (var i = 0; i < window.gaIdList.length; i++) {
     disableStr = 'ga-disable-' + window.gaIdList[i];
-    document.cookie = disableStr + '=true; expires=Thu, 31 Dec 2099 23:59:59 UTC; path=/';
+    document.cookie = disableStr + '=true; expires=Thu, 31 Dec 2099 23:59:59 UTC; path=/; domain=' + core_standards_opt_out.cookie_domain;
     window[disableStr] = true;
   }
 }
--- a/assets/scripts/google-analytics-opt-out.js
+++ b/assets/scripts/google-analytics-opt-out.js
@@ -12,9 +12,10 @@ for (var i = 0; i < window.gaIdList.length; i++) {
  */
 function gaOptout() {
   var disableStr;
+  var cookieDomain = typeof core_standards_opt_out !== 'undefined' && typeof core_standards_opt_out.cookie_domain !== 'undefined' ? core_standards_opt_out.cookie_domain : '';
   for (var i = 0; i < window.gaIdList.length; i++) {
     disableStr = 'ga-disable-' + window.gaIdList[i];
-    document.cookie = disableStr + '=true; expires=Thu, 31 Dec 2099 23:59:59 UTC; path=/; domain=' + core_standards_opt_out.cookie_domain;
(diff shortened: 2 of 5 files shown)


rodrigoaguilera:
Continued work from
https://github.com/netzstrategen/wordpress-core-standards/pull/18/files

It had an issue about setting the prefix dot in the backend and the frontend.

juanlopez4691 (approved):
Looks good to me 👍

ChristophKnopf:
I have some questions about this:
The actual solution places a cookie like `subdomain.sld.tld` right? The goal should be `.sld.tld` or did I miss something?
The above cookie should be placed regardless on which sub domain it was enabled.
Why is it necessary to add the COOKIE_DOMAIN to wp-config or just needed for testing on the test environment?

rodrigoaguilera:
> The actual solution places a cookie like subdomain.sld.tld right?

Exactly

> The goal should be .sld.tld or did I miss something?

That is what this change introduces

>  The above cookie should be placed regardless on which sub domain it was enabled.

I am not sure I understand. The cookie will be set on www.bnn.de (for example) and it can be read by jobs.bnn.de because it was set with a domain (.bnn.de) different than the usual

>  Why is it necessary to add the COOKIE_DOMAIN to wp-config or just needed for testing on the test environment?

To have this configurable was request by @sun in the previous PR
https://github.com/netzstrategen/wordpress-core-standards/pull/18#discussion_r225527814
I think is better to set this config variable than to automatically guess the domain the cookie should be set for.

sun:
> Why is it necessary to add the COOKIE_DOMAIN to wp-config or just needed for testing on the test environment?

The constant `COOKIE_DOMAIN` is a default constant of WordPress Core; see https://codex.wordpress.org/Editing_wp-config.php#Set_Cookie_Domain

By default it is set to the hostname of the site, unless a manual configuration override value is placed into wp-config.php.

This structure is a common configuration concept of most web applications.

sun:
@rodrigoaguilera The remaining todo here was to double-check that the suggested changes are actually working correctly in the following situations:

- when no GA ID has been configured
- when visiting any page
- when visiting the page containing the opt-out checkbox (e.g., imprint)
- when toggling the opt-out checkbox

…and verify that the other existing JavaScript functionality (especially concerning JS settings) of the core-standards plugin still works correctly.

Did you already do that?

rodrigoaguilera:
I checked that the cookie is set for the proper domain.

What should happen in those scenarios that you specify?
Shouldn't that be described on the ticket?

sun:
The cookie should be set and unset for the defined cookie domain.  It should still work correctly when the constant is not defined in wp-config.php.

All other pages and scenarios should still work correctly as-is; in particular, there should not be any JavaScript errors in the browser console.

rodrigoaguilera:
If you don't set COOKIE_DOMAIN then the cookie is set only for the current subdomain.
When there is no GA ID it sets no cookie.

I am not aware of any checkbox for opting out, only the link at https://stage.bnn.de/datenschutzerklaerung

It does not break the JS in in any case with errors.

assets/scripts/enable-opt-out.js
```diff
@@ -5,6 +5,6 @@
   $('.core-standards-opt-out-link').on('click', function (e) {
     e.preventDefault();
     gaOptout();
```
sun, line 7: Not caused by this PR: The toggle link does not allow to undo the operation (re-enable the tracking after opting out). This should be fixed in the future.

sun:
For completeness: When not defining a COOKIE_DOMAIN in wp-config.php (the default on every site using core-standards plugin), then the page HTML contains:
```js
var core_standards_opt_out = {"cookie_domain":""};
```
…which is then used 1:1:
```js
    document.cookie = i + "=true; expires=Thu, 31 Dec 2099 23:59:59 UTC; path=/; domain=" + core_standards_opt_out.cookie_domain,
```
…resulting in:
```js
    document.cookie = i + "=true; expires=Thu, 31 Dec 2099 23:59:59 UTC; path=/; domain="
```

This does not seem to prevent the cookie from being stored, at least in Google Chrome, so it appears to be okay to keep as it is now.

However, the empty cookie_domain inline JavaScript setting is output on all pages, which is unnecessary data that could/should be avoided. Added a fixup for this in 3f6c41f

sun (approved):

PR #25, opened 2019-01-31
Added minimal composer.json file.

rodrigoaguilera:
### Ticket
- https://app.asana.com/0/30975952354191/999106472504087

### Description
- We need a composer file in the repos we want to use as git submodules

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,20 @@
+{
+    "name": "netztrategen/core-standards",
+    "type": "wordpress-plugin",
+    "authors": [
+        {
+            "name": "Netzstrategen",
+            "email": "hallo@netzstrategen.com",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -0,0 +1,20 @@
...
+    ],
+    "dist": {
+        "url": "git@github.com:netzstrategen/wordpress-core-standards.git",
+        "type": "git"
+    },
+    "require": {
+        "php": ">=7.0.0",
+        "composer/installers": "^1"
+    }
+
```
LucaPipolo: Is this empty line a mistake? 🤔

LucaPipolo (approved):

composer.json
rodrigoaguilera: yes

PR #28, opened 2019-03-19
Fixed fast 404 rewrite rules for uploads folder cause Optimus webp images to not get served.

juliquiron:
### Ticket
- [Performance: WebP images are not delivered](https://app.asana.com/0/809933051638353/1108613656441359)

### Description
- The rewritte rules in the `.htaccess` in `wp-content/uploads` prevent any rewrite rule from main `.htaccess` to be executed. Moving it to the main one we get both functionalities.


conf/.htaccess.uploads.fast404
```diff
@@ -2,8 +2,9 @@
 # WordPress "File not found" page rendering.
```
sun: Major: If this snippet is placed into the main .htaccess file now, then developers need to be informed that they must not touch the lines as they would be duplicated otherwise.

The common way to do this in WordPress is to start the leading comment with `BEGIN` and add a trailing comment after the snippet/block starting with `END`.

conf/.htaccess.uploads.fast404
```diff
@@ -2,8 +2,9 @@
...
 <IfModule mod_rewrite.c>
   RewriteEngine On
-  RewriteCond %{REQUEST_FILENAME} !-f
-  RewriteCond %{REQUEST_FILENAME} !-d
-  RewriteCond %{REQUEST_FILENAME} !-l
-  RewriteRule ^.*$ - [nocase,R=404,L]
+  RewriteCond %{REQUEST_URI} ^(/wp-content/uploads).+$
```
sun: Major: This hardcodes the path `wp-content/uploads`, but the path can be configured / different per site.

conf/.htaccess.uploads.fast404
```diff
@@ -2,8 +2,9 @@
...
+  RewriteCond %{REQUEST_URI} !-f
+  RewriteCond %{REQUEST_URI} !-d
+  RewriteCond %{REQUEST_URI} !-l
+  RewriteRule ^.*$ - [nocase,R=404,L,NS]
```
sun: Unclear: Why is the NS flag needed?

src/Schema.php
```diff
@@ -18,9 +18,11 @@ class Schema {
   public static function activate() {
     Admin::addAccessCapability();
 
-    // Ensures arbitrary script execution protection and fast 404 responses for
-    // missing files in uploads folder.
+    // Ensures arbitrary script execution protection.
     static::ensureUploadsHtaccess();
+
+    // Fast 404 responses for missing files in uploads folder.
+    static::ensure404FastResponse();
```
```suggestion
    static::ensureFast404Response();
```

src/Schema.php
```diff
@@ -62,6 +62,17 @@ public static function ensureUploadsHtaccess() {
     static::createOrPrependFile($pathname, $template);
   }
 
+  /**
+   * Ensures fast 404 responses for missing files in uploads folder.
+   */
+  public static function ensure404FastResponse() {
```
```suggestion
  public static function ensureFast404Response() {
```

sun (changes requested):

juliquiron:
Al fixes pushed! ping for new review!

src/Schema.php
```diff
@@ -66,10 +66,14 @@ public static function ensureUploadsHtaccess() {
    * Ensures fast 404 responses for missing files in uploads folder.
    */
   public static function ensureFast404Response() {
+    $uploads_dir = wp_upload_dir(NULL, FALSE)['basedir'];
```
juanlopez4691: IMO you could use `wp_get_upload_dir()` here; it does exactly the same as your code. See https://developer.wordpress.org/reference/functions/wp_get_upload_dir/
```diff
- $uploads_dir = wp_upload_dir(NULL, FALSE)['basedir'];
+ $uploads_dir = wp_get_upload_dir()['basedir'];
```

conf/.htaccess.uploads.fast404
juliquiron: I thought that for sub-request we don't have the performance issue we are trying to fix as long the subrequest 404 are not handled by WordPress. So to only achieve one thing in this rule reducing side effects.

conf/.htaccess.uploads.fast404
```diff
@@ -1,9 +1,13 @@
-# Fast 404 error responses for missing uploaded files, skipping
-# WordPress "File not found" page rendering.
+# BEGIN core-standards
```
sun:
```suggestion
# BEGIN core-standards:fast404
```

conf/.htaccess.uploads.fast404
```diff
@@ -1,9 +1,13 @@
...
+
+  # Fast 404 error responses for missing uploaded files, skipping
+  # WordPress "File not found" page rendering.
+  RewriteCond %{REQUEST_URI} ^(UPLOAD_DIR_REPLACE).+$
+  RewriteCond %{REQUEST_URI} !-f
+  RewriteCond %{REQUEST_URI} !-d
+  RewriteCond %{REQUEST_URI} !-l
+  RewriteRule ^.*$ - [nocase,R=404,L,NS]
 </IfModule>
+# END core-standards
```
sun:
```suggestion
# END core-standards:fast404
```

conf/.htaccess.uploads.fast404
```diff
@@ -1,9 +1,13 @@
...
 <IfModule mod_rewrite.c>
   RewriteEngine On
-  RewriteCond %{REQUEST_FILENAME} !-f
-  RewriteCond %{REQUEST_FILENAME} !-d
-  RewriteCond %{REQUEST_FILENAME} !-l
-  RewriteRule ^.*$ - [nocase,R=404,L]
+
+  # Fast 404 error responses for missing uploaded files, skipping
+  # WordPress "File not found" page rendering.
+  RewriteCond %{REQUEST_URI} ^(UPLOAD_DIR_REPLACE).+$
```
sun:
```suggestion
  RewriteEngine On
  RewriteCond %{REQUEST_URI} ^(UPLOAD_DIR_REPLACE).+$
```

conf/.htaccess.uploads.fast404
```diff
@@ -1,9 +1,13 @@
-# Fast 404 error responses for missing uploaded files, skipping
-# WordPress "File not found" page rendering.
+# BEGIN core-standards
 <IfModule mod_rewrite.c>
   RewriteEngine On
```
sun:
```suggestion
```

conf/.htaccess.uploads.fast404
```diff
@@ -1,9 +1,13 @@
...
-  RewriteCond %{REQUEST_FILENAME} !-f
-  RewriteCond %{REQUEST_FILENAME} !-d
-  RewriteCond %{REQUEST_FILENAME} !-l
-  RewriteRule ^.*$ - [nocase,R=404,L]
+
```
sun:
```suggestion
```

conf/.htaccess.uploads.fast404
```diff
@@ -1,9 +1,13 @@
...
+  # Fast 404 error responses for missing uploaded files, skipping
+  # WordPress "File not found" page rendering.
+  RewriteCond %{REQUEST_URI} ^(UPLOAD_DIR_REPLACE).+$
+  RewriteCond %{REQUEST_URI} !-f
+  RewriteCond %{REQUEST_URI} !-d
+  RewriteCond %{REQUEST_URI} !-l
+  RewriteRule ^.*$ - [nocase,R=404,L,NS]
```
sun:
```suggestion
  RewriteRule ^.*$ - [nocase,nosubreq,R=404,L]
```

juanlopez4691 (approved):

sun (approved):
👍 Thanks!  Please squash-merge using the new PR title as commit message.

How are we going to remove the existing directives from `wp-content/uploads/.htaccess` upon updating the plugin?  Can you follow the same methodology and create a function `Schema::removeFromFile($pathname, $template)`, passing the old lines to remove?

juliquiron:
About removing it programmatically... I thought about it, but my concern is that it can be confusing if the files have been changed on each installation. As long the duplicate rules don't generate any specific trouble to me makes more sense that each dev/admin of the site takes care of it based on its context.

sun:
> my concern is that it can be confusing if the files have been changed on each installation

The .htaccess files are changing anyway upon updating the plugin, because this PR introduces new directives for the root one.

> As long the duplicate rules don't generate any specific trouble

They do create trouble – it's what this PR intends to fix…?

sun (changes requested):
We also need to increase version number.

juliquiron:
> They do create trouble – it's what this PR intends to fix…?

only if there are rewrite rules for the uploaded files

sun:
> only if there are rewrite rules for the uploaded files

To my knowledge, the Optimus plugin is installed on all of our sites, so we should expect rewrite rules for uploaded files on all of them.

src/Schema.php
```diff
@@ -62,6 +62,22 @@ public static function ensureUploadsHtaccess() {
...
+    $uploads_dir = wp_get_upload_dir()['basedir'];
+    $uploads_dir_relative = substr($uploads_dir, strlen(ABSPATH));
+
+    // Changes to .htaccess need to be performed in separate steps for each
+    // chunk of content that needs to be ensured. Otherwise the existing chunks
+    // would be duplicated.
+    $template = file_get_contents(Plugin::getBasePath() . '/conf/.htaccess.uploads.fast404');
+    $template = str_replace('UPLOAD_DIR_REPLACE', $uploads_dir_relative, $template);
+    static::createOrPrependFile(ABSPATH . '.htaccess', $template, "\n");
+    static::removeFromFile($uploads_dir . '/.htaccess', $template);
```
sun, line 79: `$template` needs to be the original snippet from `conf/.htaccess.uploads.fast404` from before this PR.

I would recommend to copy the original file `conf/.htaccess.uploads.fast404` into a subdirectory `conf/remove/.htaccess.uploads.fast404`

sun (approved):
👍 Did you test the plugin update manually?

juliquiron:
Yes, and works fine. So if also @juanlopez4691 is ok let's merge it.

PR #29, opened 2019-04-12
Added git reference as version to styles and scripts to bust cache.

juliquiron:
### Ticket
- [Performance: Client-side caching for asset and image files (Expires/Cache-Control)](https://app.asana.com/0/809933051638353/912303973823321)

### Description
-

### DIFF
--- a/src/Asset.php
+++ b/src/Asset.php
@@ -62,6 +62,9 @@ class Asset {
       }
       self::$gitVersion = substr($ref, 0, 8);
     }
+    else {
+      self::$gitVersion = 0;
+    }
 
     return self::$gitVersion;
   }
--- a/src/Asset.php
+++ b/src/Asset.php
@@ -1,0 +1,57 @@
+<?php
+
+/**
+ * @file
+ * Contains \Netzstrategen\CoreStandards\Asset.
+ */
+
+namespace Netzstrategen\CoreStandards;
+
+/**
+ * Administrative back-end functionality.
+ */
+class Asset {
+
+  /**
+   * Replaces default version to Git reference in script URL.
+   *
+   * @return string
+   *   The tag to print with the version updated.
+   */
+  public static function loader_tag($tag, $handle, $src) {
+    $wp_scripts = wp_scripts();
+    $default_version = $wp_scripts->default_version;
+
+    return str_replace('?ver=' . $default_version, '?ver=' . self::getGitCommitHash(), $tag);
+  }
+
+  /**
+   * Gets the reference of the last commit.
+   *
+   * Helper function to handle asset version with wp_enqueue_*
+   * functions to bust caches.
+   *
+   * @see https://developer.wordpress.org/reference/functions/wp_enqueue_script/
+   *
+   * @return string|null
+   *   The first 8 characters of the git reference.
+   */
+  public static function getGitCommitHash() {
+    $git_version = NULL;
+    if (is_dir(ABSPATH . '.git')) {
+      $ref = trim(file_get_contents(ABSPATH . '.git/HEAD'));
+      if (strpos($ref, 'ref:') === 0) {
+        $ref = substr($ref, 5);
+        if (file_exists(ABSPATH . '.git/' . $ref)) {
+          $ref = trim(file_get_contents(ABSPATH . '.git/' . $ref));
+        }
+        else {
+          $ref = substr($ref, 11);
+        }
+      }
+      $git_version = substr($ref, 0, 8);
+    }
+    return $git_version;
+  }
+
+}
--- a/src/Asset.php
+++ b/src/Asset.php
@@ -29,7 +29,10 @@ class Asset {
     $wp_scripts = wp_scripts();
     $default_version = $wp_scripts->default_version;
 
-    return str_replace('?ver=' . $default_version, '?ver=' . self::getGitCommitHash(), $tag);
+    if ($git_version = self::getGitCommitHash()) {
+      $tag = str_replace('?ver=' . $default_version, '?ver=' . $git_version, $tag);
+    }
+    return $tag;
   }
 
   /**
--- a/src/Asset.php
+++ b/src/Asset.php
@@ -8,9 +8,16 @@
 namespace Netzstrategen\CoreStandards;
 
 /**
- * Administrative back-end functionality.
+ * Asset-handling functionality.
  */
 class Asset {
+
+  /**
+   * The current commit hash.
+   *
+   * @var string
+   */
+  private static $gitVersion = NULL;
 
   /**
    * Replaces default version to Git reference in script URL.
@@ -37,7 +44,11 @@ class Asset {
    *   The first 8 characters of the git reference.
    */
   public static function getGitCommitHash() {
-    $git_version = NULL;
+    // Reuse the value if already calculated.
+    if (self::$gitVersion != NULL) {
+      return self::$gitVersion;
+    }
+
     if (is_dir(ABSPATH . '.git')) {
       $ref = trim(file_get_contents(ABSPATH . '.git/HEAD'));
       if (strpos($ref, 'ref:') === 0) {
@@ -49,9 +60,10 @@ class Asset {
           $ref = substr($ref, 11);
         }
       }
-      $git_version = substr($ref, 0, 8);
+      self::$gitVersion = substr($ref, 0, 8);
     }
-    return $git_version;
+
+    return self::$gitVersion;
   }
 
 }
--- a/src/Asset.php
+++ b/src/Asset.php
@@ -45,7 +45,7 @@ class Asset {
    */
   public static function getGitCommitHash() {
     // Reuse the value if already calculated.
-    if (self::$gitVersion != NULL) {
+    if (isset(self::$gitVersion)) {
       return self::$gitVersion;
     }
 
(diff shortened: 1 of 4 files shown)


src/Plugin.php
```diff
@@ -198,6 +198,35 @@ public static function getBasePath() {
...
+   *
+   * Helper function to handle asset version with wp_enqueue_*
+   * functions to bust caches.
+   *
+   * @see https://developer.wordpress.org/reference/functions/wp_enqueue_script/
+   *
+   * @return string|null
+   *   The first 8 characters of the git reference.
+   */
+  public static function getGitRef() {
```
```suggestion
  public static function getGitCommitHash() {
```

src/TrackingOptOut.php
```diff
@@ -106,7 +106,7 @@ public static function renderSettings() {
   }
 
   public static function enqueueScriptsFirstHead() {
-    wp_enqueue_script(Plugin::PREFIX . '-google_analytics_opt_out', Plugin::getBaseUrl() . '/dist/scripts/google-analytics-opt-out.js');
+    wp_enqueue_script(Plugin::PREFIX . '-google_analytics_opt_out', Plugin::getBaseUrl() . '/dist/scripts/google-analytics-opt-out.js', FALSE, Plugin::getGitRef());
```
sun: The manual approach is prone to errors and mistakes and won't catch all asset files needing a version. Any chance to apply this on a more global scope for all scripts/styles that do not state a version yet?

Are the enqueue API functions invoking any hooks? (or alternatively the rendering functions)

src/Plugin.php
sun: This feature should be moved into a separate class/file within Core Standards, so that it is properly encapsulated. For example, class `Asset`

sun (changes requested):

src/TrackingOptOut.php
juliquiron: I do not agree. Some styles/scripts will depend on the git version (project specific theme and custom module) others on the plugin version (basically all plugins), others in the WordPress version. Handle it with a generic filter on render or process will require to handle this logic in a centralized place (or implement it in different places for each case, so again not centralized). IMHO makes more sense that each plugin or theme is aware and responsible to handle the cache invalidation of its styles and scripts.

sun: Hence the suggestion to only apply it automatically to all assets that do not supply a version yet. The current Git commit hash always works as a safe fallback to guarantee that all assets have one. Any other specialized version hash only improves the invalidation frequency (to not invalidate automatically with every deployment).

juliquiron: ok, I will rework it.

src/UserFrontend.php
```diff
@@ -90,7 +90,7 @@ public static function login_form_defaults(array $args) {
    */
   public static function login_enqueue_scripts() {
     if (file_exists(get_stylesheet_directory() . '/dist/styles/login.css')) {
-      wp_enqueue_style('theme/login', get_stylesheet_directory_uri() . '/dist/styles/login.css');
+      wp_enqueue_style('theme/login', get_stylesheet_directory_uri() . '/dist/styles/login.css', FALSE);
```
sun: (and elsewhere) The 3rd argument to https://developer.wordpress.org/reference/functions/wp_enqueue_style/ is an array of dependencies. The 4th argument would be the version, but it already defaults to `FALSE`, so no need to specify it manually.

src/Plugin.php
```diff
@@ -198,6 +201,48 @@ public static function getBasePath() {
     return dirname(__DIR__);
   }
 
+  /**
+   * Replaces default version to Git reference in script URL.
+   *
+   * @return string
+   *   The tag to print with the version updated.
+   */
+  public static function script_loader_tag($tag, $handle, $src) {
```
sun: Since core-standards is a collection of many different adjustments and features, we should strive for a clear code organization of features, ideally one feature or topic per class.  Can we move the callbacks (not their registration) into a new class `Asset`?

sun (changes requested):

src/Asset.php
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * @file
+ * Contains \Netzstrategen\CoreStandards\Asset.
+ */
+
+namespace Netzstrategen\CoreStandards;
+
+/**
+ * Administrative back-end functionality.
```
sun:
```suggestion
 * Asset-handling functionality.
```

src/Asset.php
```diff
@@ -0,0 +1,57 @@
...
+   * Helper function to handle asset version with wp_enqueue_*
+   * functions to bust caches.
+   *
+   * @see https://developer.wordpress.org/reference/functions/wp_enqueue_script/
+   *
+   * @return string|null
+   *   The first 8 characters of the git reference.
+   */
+  public static function getGitCommitHash() {
+    $git_version = NULL;
```
sun: This value should be statically cached on the class to only execute it once per request.

src/TrackingOptOut.php
```diff
@@ -106,7 +106,7 @@ public static function renderSettings() {
   }
 
   public static function enqueueScriptsFirstHead() {
-    wp_enqueue_script(Plugin::PREFIX . '-google_analytics_opt_out', Plugin::getBaseUrl() . '/dist/scripts/google-analytics-opt-out.js');
+    wp_enqueue_script(Plugin::PREFIX . '-google_analytics_opt_out', Plugin::getBaseUrl() . '/dist/scripts/google-analytics-opt-out.js', FALSE);
```
sun: (and elsewhere) The 3rd argument to https://developer.wordpress.org/reference/functions/wp_enqueue_script/ is an array of dependencies, so adding a Boolean seems wrong here.

sun (changes requested):

src/Asset.php
```diff
@@ -0,0 +1,69 @@
...
+ * Asset-handling functionality.
+ */
+class Asset {
+
+  /**
+   * The current commit hash.
+   *
+   * @var string
+   */
+  private static $gitVersion = NULL;
```
sun:
```suggestion
  private static $gitVersion;
```

src/Asset.php
```diff
@@ -0,0 +1,69 @@
...
+   * functions to bust caches.
+   *
+   * @see https://developer.wordpress.org/reference/functions/wp_enqueue_script/
+   *
+   * @return string|null
+   *   The first 8 characters of the git reference.
+   */
+  public static function getGitCommitHash() {
+    // Reuse the value if already calculated.
+    if (self::$gitVersion != NULL) {
```
sun:
```suggestion
    if (isset(self::$gitVersion)) {
```

src/Asset.php
```diff
@@ -0,0 +1,69 @@
...
+        $ref = substr($ref, 5);
+        if (file_exists(ABSPATH . '.git/' . $ref)) {
+          $ref = trim(file_get_contents(ABSPATH . '.git/' . $ref));
+        }
+        else {
+          $ref = substr($ref, 11);
+        }
+      }
+      self::$gitVersion = substr($ref, 0, 8);
+    }
```
sun, line 67:
```suggestion
    }
    else {
      self::$gitVersion = 0;
    }
```

src/Asset.php
```diff
@@ -0,0 +1,69 @@
...
+   * Replaces default version to Git reference in script URL.
+   *
+   * @return string
+   *   The tag to print with the version updated.
+   */
+  public static function loader_tag($tag, $handle, $src) {
+    $wp_scripts = wp_scripts();
+    $default_version = $wp_scripts->default_version;
+
+    return str_replace('?ver=' . $default_version, '?ver=' . self::getGitCommitHash(), $tag);
```
sun:
```suggestion
    if ($git_version = self::getGitCommitHash()) {
      $tag = str_replace('?ver=' . $default_version, '?ver=' . $git_version, $tag);
    }
    return $tag;
```

src/TrackingOptOut.php
```diff
@@ -116,7 +116,7 @@ public static function enqueueScriptsFirstHead() {
    * @implements wp_enqueue_scripts.
    */
   public static function wp_enqueue_scripts() {
-    wp_enqueue_script(Plugin::PREFIX . '-enable-opt_out', Plugin::getBaseUrl() . '/dist/scripts/enable-opt-out.js', ['jquery'], FALSE, TRUE);
+    wp_enqueue_script(Plugin::PREFIX . '-enable-opt_out', Plugin::getBaseUrl() . '/dist/scripts/enable-opt-out.js', ['jquery'], TRUE);
```
sun: Unless I'm mistaken, the intention here is to load the opt-out script after the content has been loaded, so that it can be applied to the ga-optout shortcode in the content body.

Did you check whether that still works correctly when loading the script in the header?

`TRUE` is not a supported value for the `$ver` argument of https://developer.wordpress.org/reference/functions/wp_enqueue_script/

I would think that the current value `FALSE` was intended:
> If version is set to false, a version number is automatically added equal to current installed WordPress version. 

```suggestion
    wp_enqueue_script(Plugin::PREFIX . '-enable-opt_out', Plugin::getBaseUrl() . '/dist/scripts/enable-opt-out.js', ['jquery'], FALSE, TRUE);
```

sun (approved):
👍 Thanks!

PR #30, opened 2019-05-15
Added constant allowing to disable cookie notice via wp-config.php.

sun:
### Ticket
- [Deactivate Cookie notice in Core standards](https://app.asana.com/0/32260443473239/1122794580966854/f)

### Description
- Normally the cookie notice is enabled by default, but in edge cases a project might not need or want it.  The new constant allows to disable it via `wp-config.php` by setting:
```php
const CORE_STANDARDS_COOKIE_NOTICE = FALSE;
```

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -90,7 +90,7 @@ class Plugin {
     // Output client-side cookie notice on all pages.
     // Use a slightly higher weight to prevent the bar being output
     // before other footer content.
-    if (!defined('CORE_STANDARDS_COOKIE_NOTICE') || CORE_STANDARDS_COOKIE_NOTICE === FALSE) {
+    if (!defined('CORE_STANDARDS_COOKIE_NOTICE') || CORE_STANDARDS_COOKIE_NOTICE !== FALSE) {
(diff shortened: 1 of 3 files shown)


src/Plugin.php
```diff
@@ -90,7 +90,9 @@ public static function init() {
     // Output client-side cookie notice on all pages.
     // Use a slightly higher weight to prevent the bar being output
     // before other footer content.
-    add_action('wp_footer', __CLASS__ . '::wp_footer', 12);
+    if (!defined('CORE_STANDARDS_COOKIE_NOTICE') || CORE_STANDARDS_COOKIE_NOTICE === FALSE) {
```
fabianmarz: Your code will prevent the cookie notice from rendering if the constant set and `TRUE`.

```suggestion
    if (!defined('CORE_STANDARDS_COOKIE_NOTICE') || CORE_STANDARDS_COOKIE_NOTICE !== FALSE) {
```

fabianmarz (approved):

juanlopez4691 (approved):
LGTM! 👍

PR #31, opened 2019-05-22
Changed cookie notice styling to be more generic.

colourgarden:
### Ticket
- [core-standards: Improve UX and UI of cookie notice](https://app.asana.com/0/30975952354191/1123925021923421/f)


juanlopez4691 (approved):
@colourgarden LGTM. My only concern is that IMO the "Mehr infos" link could use some text decoration/underline or similar visual highlighting. It doesn't look clearly as a link, to me.

colourgarden:
Good spot. On FECO (where I tested it), the links are underlined by default 🙈 
`text-decoration: underline` added to the link.

sun:
The previous code assumed that it would get overridden with more specific styles by the theme.  With this change, the amount of markup and styles being output inline on every page becomes significantly larger.  How can we achieve the optimal performance (in particular no temporary visibility) for first-time visitors as well as visitors accessing multiple pages?

colourgarden:
> The previous code assumed that it would get overridden with more specific styles by the theme.

That's what we want to avoid. We want to make the cookie bar more generic (and improve UX) to avoid having to style it for every client.

> With this change, the amount of markup and styles being output inline on every page becomes significantly larger.

This is true although remember that we will also lose 20-30 lines of (admittedly cached) custom CSS in each theme. We could certainly move the CSS into it's own stylesheet so it gets cached?

> in particular no temporary visibility

My assumption is that because we're using localStorage, the cookie notice isn't hidden by default so that users on older browsers still see it. localStorage has support way back to IE8 but I guess, as this is a legal requirement, we have to ensure 100% of visitors see the notice, not just 99.9%. Unsure.

The JS is being executed immediately after the markup is read so I don't think FOUC is a big concern. Again, we could move it into it's own JS file so it gets cached?

> visitors accessing multiple pages

Are you referring to caching here?

sun:
Yes, the cookie notice is displayed by default and hidden via JS based on LocalStorage (and not a cookie for compatibility with reverse proxies).  The CSS/JS is inline so that the bar is immediately hidden upon rendering.

For that reason, the JS would need to stay inline.  But the inline HTML and CSS could be reduced to the bare minimum (visibility, positioning, dimensions), moving all decorative styling (including the SVG) into a separate CSS file.  That way, we minimize the output / body size on all pages, achieve immediate execution of the base output, and cache the decorative styling across multiple page views (which is only required when visible anyway).  Does that sound sensible and doable to you?

I also just noticed that the new code is not accessible for screen-readers; i.e., the link text should be "Close" or "Accept".  CSS can replace the text with the SVG/image, but we need to follow basic markup standards.

juanlopez4691:
@colourgarden Also remember to bump plugin version accordingly.

sun:
@colourgarden and I discussed and agreed that the only important things to keep inline are the HTML and the JS (causing the HTML to be immediately hidden).  CSS is render-blocking anyway, so the whole CSS (including the SVG) can be moved into a file.

colourgarden:
@sun Ready to be reviewed again (sorry about the many fixup commits 🙈). I worked with Fabian to add the plugin version to the `<link>` to improve caching further.

The SVG is still inlined in the DOM as it changes colour on hover (via CSS `fill`). To achieve this with the SVG being embedded in the CSS file, we'd need another image. Considering the SVG code is only 400 bytes, I think this is an ok compromise for developer usability/readability.

src/Plugin.php
```diff
@@ -245,6 +245,15 @@ public static function loadTextdomain() {
     load_plugin_textdomain(static::L10N, FALSE, static::L10N . '/languages/');
   }
 
+  /**
+   * Returns the plugin meta data.
+   *
+   * @todo Implement caching.
+   */
+  public static function get_plugin_data() {
```
```suggestion
  public static function getMetaData() {
```

However, retrieving this at runtime is pretty slow; the function should normally be used in the admin only (or cached appropriately).

I do not actually think that the custom plugin version is necessary, because #29 is actually about to add the current Git commit hash as version number to all assets using the "default version" (the WP Core version) for cache busting.

Therefore, I'd recommend to simply remove this part and just ensure that it is using the default version instead (which I believe it does by default).

src/UserFrontend.php
```diff
@@ -67,6 +70,16 @@ public static function auth_cookie_expiration($duration, $user_id, $remember) {
     return $duration * 15;
   }
 
+  /**
+   * @implements wp_enqueue_scripts
+   */
+  public static function cookie_notice_enqueue_styles() {
+    $ver = Plugin::get_plugin_data()['Version'];
+    if (file_exists(Plugin::getBasePath() . '/dist/styles/cookie-notice.css')) {
```
sun: Why would this file not exist?  This condition can be removed I think.

templates/cookie-notice.php
```diff
@@ -2,35 +2,28 @@
 
 namespace Netzstrategen\CoreStandards;
 
+if (has_action('core_standards/cookie_notice/css_extend')) {
+  echo '<style>';
+  do_action('core_standards/cookie_notice/css_extend');
```
sun: Does this still make sense? — Existing styling needs to be migrated anyway to this new structure…?

gulpfile.js
```diff
@@ -12,7 +12,7 @@ const uglify = require('gulp-uglify');
 
 gulp.task('styles', () => {
   gulp.src('assets/styles/**/*.scss')
-  .pipe(sass({outputStyle: 'expanded'}).on('error', sass.logError))
+  .pipe(sass({outputStyle: 'compressed'}).on('error', sass.logError))
```
sun, line 15: This can cause merge conflicts in the future, but we can try it out for now.

sun (commented):
Thanks! Almost done, but identified some remaining issues:

src/Plugin.php
fabianmarz: I was suggesting this to @colourgarden but also mentioned we have the git versioning already but we're not using this in the core standards plugin currently, so we thought of introducing this. I doubt the git versioning will work for plugins published on the official WP plugin repository (haven't checked), so getting the plugin version should be more stable, no?

By default it'll use the current WP version which will not be sufficient to invalidate the assets.

src/UserFrontend.php
colourgarden: Removed.

templates/cookie-notice.php
colourgarden: As the action already existed before, I thought it might be helpful to leave it there.

sun (approved):

templates/cookie-notice.php
```diff
@@ -1,36 +1,23 @@
...
-</style>
-
-<div id="cookie-notice" role="banner">
-  <span><?= __('This website uses cookies to give you the best possible service. By continuing to use the website, you agree to the use of cookies.', Plugin::L10N) ?></span>
-  <a href="#" id="cookies-accept" class="btn"><?= __('OK') ?></a>
-  <a href="<?= apply_filters('core_standards/cookie_notice/policy_link', '/privacy') ?>" target="_self" class="btn"><?= __('Learn more', Plugin::L10N) ?></a>
+<div class="cookie-notice" id="cookie-notice" role="banner">
+  <div class="cookie-notice__body">
+    <?= __('This website uses cookies to give you the best possible service. By continuing to use the website, you agree to the use of cookies.', Plugin::L10N) ?>
+    <a class="cookie-notice__link" href="<?= apply_filters('core_standards/cookie_notice/policy_link', '/privacy') ?>" target="_self"><?= __('Learn more', Plugin::L10N) ?></a>
```
sun, line 9: As just discussed in person, the privacy policy link should be incorporated into the text.

Let's make sure to create a task for this and resolve it before this change gets deployed to further sites, as this would cause lots of unnecessary update work otherwise.

sun: Initial text suggestion after doing a quick review of prominent German websites:

```diff
     <?= __('This website uses cookies to give you the best possible service.
-            By continuing to use the website, you agree to the use of cookies.', Plugin::L10N) ?>
+            By continuing to use it, you agree to our <a href="%s">privacy policy</a>.', Plugin::L10N) ?>
```

_"Diese Website verwendet Cookies, um den bestmöglichen Service zu bieten. Mit der weiteren Nutzung akzeptieren Sie unsere Datenschutzrichtlinie."_

The most problematic aspect is the German translation, so the follow-up task should focus on that.

PR #32, opened 2019-05-28
Fixed cookie notice text and privacy policy link.

juanlopez4691:
### Ticket
- [core-standards: Update cookie notice text](https://app.asana.com/0/30975952354191/1124673186025759)


languages/core-standards-de_DE.po
```diff
@@ -116,3 +115,13 @@ msgstr ""
...
+
+#~ msgid ""
+#~ "This website uses cookies to give you the best possible service. By "
+#~ "continuing to use the website, you agree to the use of cookies."
+#~ msgstr ""
+#~ "Diese Website verwendet Cookies. Wenn Sie sich weiterhin auf dieser Seite "
+#~ "aufhalten, gehen wir von Ihrem Einverständnis aus."
+
+#~ msgid "Learn more"
+#~ msgstr "Mehr Infos"
```
sun: These should be removed.

templates/cookie-notice.php
```diff
@@ -5,8 +5,7 @@
 
 <div class="cookie-notice" id="cookie-notice" role="banner">
   <div class="cookie-notice__body">
-    <?= __('This website uses cookies to give you the best possible service. By continuing to use the website, you agree to the use of cookies.', Plugin::L10N) ?>
-    <a class="cookie-notice__link" href="<?= apply_filters('core_standards/cookie_notice/policy_link', '/privacy') ?>" target="_self"><?= __('Learn more', Plugin::L10N) ?></a>
+    <?= sprintf(__('To give you the best possible service, this website uses cookies. By continuing to use it, you agree to our <a href="%s" target="_self">privacy policy</a>.', Plugin::L10N), apply_filters('core_standards/cookie_notice/policy_link', '/privacy')); ?>
```
sun: The target is superfluous (_self is the default).

```suggestion
    <?= sprintf(__('To give you the best possible service, this website uses cookies. By continuing to use it, you agree to our <a href="%s">privacy policy</a>.', Plugin::L10N), apply_filters('core_standards/cookie_notice/policy_link', '/privacy')); ?>
```

The class has been removed already, but the string translations were not updated yet.

sun (changes requested):

languages/core-standards-de_DE.po
juanlopez4691: Fixed.

templates/cookie-notice.php
juanlopez4691: Ack! Sorry. Will do.

colourgarden: We can remove the class but we still need the styling applied. If we're going this route then we also need to update the CSS. [This block](https://github.com/netzstrategen/wordpress-core-standards/blob/master/assets/styles/cookie-notice.scss#L21-L33) will need to be nested inside the block above it using an `a` element selector.

juanlopez4691:
@sun Removing it makes the site styling for links to be applied and in WOHN this doesn't look good: dark red on dark background. Also, rollover colour for both BEGA/WOHN is dark grey. I restored the link class `cookie-notice__link` so that link is always legible.

sun:
> Removing it makes the site styling for links to be applied and in WOHN this doesn't look good: dark red on dark background. Also, rollover colour for both BEGA/WOHN is dark grey. I restored the link class `cookie-notice__link` so that link is always legible.

@juanlopez4691 Did you adjust the CSS as suggested in https://github.com/netzstrategen/wordpress-core-standards/pull/32#discussion_r287998666?

juanlopez4691:
> @juanlopez4691 Did you adjust the CSS as suggested in #32 (comment)?

@sun missed that comment, sorry. Fixed.

assets/styles/cookie-notice.scss
```diff
@@ -16,27 +16,22 @@
...
+      color: currentColor;
+      text-decoration: underline;
+    }
 
-  &:hover,
-  &:focus {
-    opacity: 0.8;
+    &:hover,
+    &:focus {
+      opacity: 0.8;
```
sun: Not sure why we are fiddling with the opacity here. Might also be a remnant of the former button styling that is unnecessary for a text link now.

languages/core-standards-de_DE.po
sun: Still visible here? Did you commit the change?

sun (approved):

juanlopez4691:
@sun fixed.

assets/styles/cookie-notice.scss
colourgarden: This styling is required and wasn't a leftover from the previous cookie notice.

It's to provide a hover state without relying on a particular text color. For example, if the color of the notice was changed to `#000` for a particular client site, there would still be a suitable hover state without having to specify another color.

PR #36, opened 2019-07-18
Added protection against DoS and brute-force attacks on wp-login.php and xmlrpc.php.

sun:
### Ticket
- https://netzstrategen.slack.com/archives/C03KDSK1E/p1563383725014000

### Description

- Replaces `/wp-login.php` with `/login.php` and blocks access to `/wp-login.php`.

- Updates all links and redirects in the system to point to the new front controller.

- Requires the core-standards plugin to be deactivated and activated to install the new rewrite rules in .htaccess.

- Replaces https://github.com/netzstrategen/mwk/pull/14

- Also protects against accesses to xmlrpc.php (and thus resolves 
[Performance: Ensure xmlrpc.php is blocked](https://app.asana.com/0/354170587449545/1131320151887257/f))

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -40,6 +40,18 @@ class Plugin {
   private static $baseUrl;
 
   /**
+   * @implements plugins_loaded
+   */
+  public static function plugins_loaded() {
+    // Change URLs pointing to /wp-login.php into /login.php.
+    add_filter('network_site_url', __NAMESPACE__ . '\Security::site_url', 100, 3);
+    add_filter('site_url', __NAMESPACE__ . '\Security::site_url', 100, 3);
+    add_filter('wp_redirect', __NAMESPACE__ . '\Security::wp_redirect', 1);
+
+    Plugin::loadTextdomain();
+  }
+
+  /**
    * @implements init
    */
   public static function init() {
--- a/src/Schema.php
+++ b/src/Schema.php
@@ -23,6 +23,9 @@ class Schema {
 
     // Fast 404 responses for missing files in uploads folder.
     static::ensureFast404Response();
+
+    // Rename /wp-login.php into /login.php and deny access to XML-RPC API.
+    static::ensureFrontControllerAccess();
   }
 
   /**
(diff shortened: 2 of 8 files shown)


src/Plugin.php
```diff
@@ -39,6 +39,18 @@ class Plugin {
    */
   private static $baseUrl;
 
+  /**
+   * @implements plugins_loaded
+   */
+  public static function plugins_loaded() {
+    // Change URLs pointing to /wp-login.php into /login.php.
+    add_filter('network_site_url', __NAMESPACE__ . '\Security::site_url', 100, 3);
```
fabianmarz, line 47: Wondering if this should an even higher priority than 100, e.g. 999 in case other plugins rewrite it? Even if I'm not aware of hooks rewriting the URLs in our projects we should consider this is a public plugin which may is used by other users as well.

src/Schema.php
```diff
@@ -79,6 +82,14 @@ public static function ensureFast404Response() {
     static::removeFromFile($uploads_dir . '/.htaccess', $template);
   }
 
+  /**
+   * Ensures fast 404 responses for missing files in uploads folder.
```
fabianmarz: The comment needs to be adjusted.

src/Schema.php
```diff
@@ -79,6 +82,14 @@ public static function ensureFast404Response() {
...
+   */
+  public static function ensureFrontControllerAccess() {
+    $template = file_get_contents(Plugin::getBasePath() . '/conf/.htaccess.security');
+    Schema::createOrPrependFile(ABSPATH . '.htaccess', $template, "\n");
```
fabianmarz: We should stay consistent with the class references. In [`ensureFast404Response()`](https://github.com/netzstrategen/wordpress-core-standards/pull/36/files#diff-8db9667fb5a2702c4e96d6ba0adf79bdR82) we use `static`, which is our default way of referencing methods of the same class, no?

fabianmarz (commented):
Good PR! Tested on stage.bnn.de and works correctly 👍 Just some minor things mentioned.

src/Plugin.php
sun: 100 is relatively unusual already, I think we should rather increase it when we run into issues.

sun:
Additionally added a readme.txt in order to start documenting the features. Obviously incomplete, only contains this one new feature for now; needs to be completed separately.

A new feature… reminds me that the version needs to be increased.

sun:
Increased version. Rebased onto latest master.

fabianmarz (approved):
LGTM now 👍

PR #38, opened 2019-10-14
Added modal dialog to choose cookie groups triggering events to inject scripts only after consent.

sun:
### Ticket
- [Expand cookie notice banner into modal dialog and trigger GTM via opt-in event](https://app.asana.com/0/809933051638353/1143867514894049/f)

### Description
-


templates/cookie-notice.php
```diff
@@ -1,58 +1,100 @@
...
+    <div class="form-actions">
+      <button class="button btn" data-js="confirm"><?= __('Meine Auswahl bestätigen', Plugin::L10N) ?></button>
+      <button class="button button--primary btn btn--primary" data-js="confirm-all"><?= __('Gönn ich mir alle', Plugin::L10N) ?></button>
+    </div>
   </div>
-  <!--a class="cookie-notice__close" href="#" id="cookies-accept" aria-label="<?= __('OK', Plugin::L10N) ?>"></a-->
 </div>
 
 <script>
 (function () {
```
sun: We should probably move this content into an actual .js file in assets/scripts now, so that it gets optimized. We can include it inline without parsing with `file_get_contents()` or can consider to load it separately now.

templates/cookie-notice.php
```diff
@@ -1,58 +1,100 @@
...
+
+  function hideCookieNotice() {
+    const cookie_notice = document.getElementById('cookie-notice');
+    cookie_notice.setAttribute('hidden', 'true');
+    document.body.classList.remove('has-cookie-notice');
+  }
+
+  function DOMContentLoaded() {
+    document.body.classList.add('has-cookie-notice');
+    document.addEventListener('click', (event) => {
```
sun: We should be careful with backwards-compatibility with the modal dialog now, because the user wouldn't be able to dismiss it.  But we can probably control that more granularly when compiling the script anyway.

templates/cookie-notice.php
```diff
@@ -1,58 +1,100 @@
...
+  }
+  return implode($attributes, ' ');
+}
 ?>
 
 <div class="cookie-notice" id="cookie-notice" role="banner">
   <div class="cookie-notice__body">
-    <h1>Kekse?</h1>
+    <h1><?= __('Kekse?', Plugin::L10N) ?></h1>
     <p><?= sprintf(__('Für den bestmöglichen Service bietet Ihnen diese Website Cookies. Mehr Infos finden Sie in unserer <a href="%s">Cookie-Bar</a>.', Plugin::L10N), apply_filters('core_standards/cookie_notice/policy_link', '/privacy')); ?></p>
```
sun: We still need to move the strings into translation files and replace them with English ones here.  It might be a good idea to move all string definitions into variables at the beginning of the file, to make editing easier? (They might be overridden by configuration in the future, which would move them to the beginning anyway.)

sun (commented):
👍

fabianmarz:
@sun, please check my latest commits. I replaced the `gulpfile` with our gulp task collection as we have the `babel` transpiler running there. I also added a `.stylelintrc` configuration as the `gulp-task-collection` requires one. This fixed also some SCSS errors already and will help us preventing further errors. A `.eslintrc` file was already there. I thought of adding a dedicated PR for it, but I think it's fine to rebase and use this PR for introducing the `gulp-task-collection` as well.

assets/scripts/cookie-consent.js
```diff
@@ -0,0 +1,59 @@
...
+        if (event.target.dataset.js === 'confirm-all') {
+          checkbox.checked = true;
+        }
+        data.consent[checkbox.value] = checkbox.checked;
+      }
+      window.localStorage.setItem('cookie-consent', JSON.stringify(data));
+      $.post(window.core_standards.ajaxurl, {
+        action: 'core-standards/log_cookie_consent',
+        consent: data.consent,
+        id: generate_uuid(),
```
sun: 1. Major: This posts a different UUID than the one in data to the backend.

1. The version is not posted to the backend, it should be included.  Why are we not simply posting `...data` instead of individual keys?

assets/scripts/cookie-consent.js
```diff
@@ -0,0 +1,59 @@
...
+      };
+      const checkboxes = document.querySelectorAll('input[name=cookies]');
+      for (const checkbox of checkboxes) {
+        if (event.target.dataset.js === 'confirm-all') {
+          checkbox.checked = true;
+        }
+        data.consent[checkbox.value] = checkbox.checked;
+      }
+      window.localStorage.setItem('cookie-consent', JSON.stringify(data));
+      $.post(window.core_standards.ajaxurl, {
```
sun, line 38: `ajaxurl` is too generic, should be `ajaxurl_cookie_consent`

core-standards.php
```diff
@@ -2,7 +2,7 @@
 
 /*
   Plugin Name: Core Standards
-  Version: 1.24.0
+  Version: 2.0.0
```
sun, line 5: Which incompatibility is introduced by this code?

languages/core-standards-de_DE_formal.po
```diff
@@ -85,19 +85,51 @@ msgstr ""
...
+msgid "Chocolate cookies"
+msgstr "Cookies für Statistiken"
+
+#: templates/cookie-consent.php:18
+msgid "Statistics"
+msgstr "Statistiken"
+
+#: templates/cookie-consent.php:22
+msgid "Space cookies"
+msgstr "Cookies zur Personalisierung"
```
sun: Major: The English words in the source code needs to contain the formal / neutral / regular words. The "funny" version must only be contained in the informal German translation.

We will most likely remove the funny version at some point, therefore the translations should not be based on it but the regular language instead.

src/Logger.php
```diff
@@ -0,0 +1,40 @@
...
+ */
+class Logger {
+
+  /**
+   * Returns cookie consent log file path.
+   *
+   * @return string
+   */
+  public static function getLogFile(): string {
+    $log_file = apply_filters('core-standards/cookie-consent/logfile', wp_upload_dir()['path'] . '/cookie-consent.log');
```
sun: Major: The file can be publicly accessed, which must not be possible.  We need to move it into a subdirectory and protected that folder by putting a .htaccess file into it.  I believe the core-standards plugin already contains functions to do.

We need to ensure that the folder is group-writable, and that the .htaccess file contains `Require all denied`

src/Logger.php
```diff
@@ -0,0 +1,40 @@
...
+   * Writes the given message into the cookie consent log file.
+   *
+   * @param array $message
+   *   The message to write into the log file.
+   */
+  public static function writelog(array $message) {
+    $log_file = static::getLogFile();
+    $file_handle = fopen($log_file, 'a');
+    fwrite($file_handle, json_encode($message) . "\r\n");
+    fclose($file_handle);
```
sun: Rules of thumb:
- Use `file_put_contents()` unless you have a good reason not to. Use the `FILE_APPEND` constant when desired.
- Always specify options to `json_encode()`.
```diff
-    $file_handle = fopen($log_file, 'a');
-    fwrite($file_handle, json_encode($message) . "\r\n");
-    fclose($file_handle);
+    $message = json_encode($message, JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE);
+    file_put_contents($log_file, $message . "\n", FILE_APPEND);
```

src/Plugin.php
```diff
@@ -234,6 +239,35 @@ public static function xmlrpc_methods($methods) {
...
+  /**
+   * Logs cookie consent message.
+   */
+  public static function log_cookie_consent() {
+    $consent = $_POST['consent'] ?? [];
+    $data = [
+      'consent' => $consent,
+      'timestamp' => current_time('timestamp'),
+      'user_agent' => $_SERVER['HTTP_USER_AGENT'] ?? '',
+      'version' => CORE_STANDARDS_COOKIE_CONSENT_VERSION,
```
sun: As mentioned before, I think the version should be passed from JS.

src/Plugin.php
```diff
@@ -234,6 +239,35 @@ public static function xmlrpc_methods($methods) {
...
+      'placed_url' => wp_get_raw_referer(),
```
sun: 1. Major: The HTTP referrer is not reliable. This value should be passed from JS, it's trivial to generate.

1. Rename the key to "url".

src/Plugin.php
```diff
@@ -234,6 +239,35 @@ public static function xmlrpc_methods($methods) {
...
+  }
+
+  /**
+   * Logs cookie consent message.
+   */
+  public static function log_cookie_consent() {
+    $consent = $_POST['consent'] ?? [];
+    $data = [
+      'consent' => $consent,
+      'timestamp' => current_time('timestamp'),
```
sun: This should be the first element in the logged data.

Why are we logging machine values that cannot be read?  I'd suggest to change the value to `date_i18n('r')` or `date_i18n('c')` instead.

src/Plugin.php
```diff
@@ -234,6 +239,35 @@ public static function xmlrpc_methods($methods) {
...
+      'consent_version' => CORE_STANDARDS_COOKIE_CONSENT_VERSION,
+    ];
+    echo '<script>var core_standards = ' . json_encode($data) . ';</script>';
+  }
+
+  /**
+   * Logs cookie consent message.
+   */
+  public static function log_cookie_consent() {
+    $consent = $_POST['consent'] ?? [];
```
sun: The consent data should never be empty.  We should log an error instead and return early if it is.

src/Plugin.php
```diff
@@ -234,6 +239,35 @@ public static function xmlrpc_methods($methods) {
...
+    $data = [
+      'consent' => $consent,
+      'timestamp' => current_time('timestamp'),
+      'user_agent' => $_SERVER['HTTP_USER_AGENT'] ?? '',
+      'version' => CORE_STANDARDS_COOKIE_CONSENT_VERSION,
+      'placed_url' => wp_get_raw_referer(),
+      'user' => get_current_user_id(),
+      'id' => uniqid(rand(), TRUE),
```
sun: Major: This generates yet another ID that doesn't match the ID in the stored cookie.

src/Plugin.php
```diff
@@ -234,6 +239,35 @@ public static function xmlrpc_methods($methods) {
...
 
+  /**
+   * @implements wp_head
+   */
+  public static function wp_head() {
+    $data = [
+      'ajaxurl' => admin_url('admin-ajax.php'),
+      'consent_version' => CORE_STANDARDS_COOKIE_CONSENT_VERSION,
+    ];
+    echo '<script>var core_standards = ' . json_encode($data) . ';</script>';
```
sun: 1. WP has a built-in support for JavaScript variables via wp_enqueue_scripts() or a closely related function. This should be used instead.

1. Please make sure that this JS variable name is not taken yet. If I'm not mistaken, core-standards also uses gettext string translations for JavaScript, which are potentially output with the same name.

1. Options for json_encode(), as mentioned earlier in this PR.

1. Major: The constant `CORE_STANDARDS_COOKIE_CONSENT_VERSION` is not defined anywhere.

1. All of the added code in this file, including the constant definition, should be moved to a new CookieConsent class. As the code is executed on all pages, you can simply move the code as-is and create a `CookieConsent::init()` method, which you call from `Plugin::init()`.

sun (commented):
🔴 Unfortunately, this PR still needs quite some work. I found several major problems in my review.

core-standards.php
evemf: @fabianmarz may need to answer that

assets/scripts/cookie-consent.js
fabianmarz: Why is this too generic if it's in the `core_standards` scope? If another feature of `core-standards` needs to use the admin AJAX URL we need to add another key with the same value? 🤔

core-standards.php
fabianmarz: The cookie consent changes from a simple banner to an interactive modal window and thus can change the expected user behaviour.

src/Plugin.php
fabianmarz: 1. Not sure what you mean by this. Do you mean https://developer.wordpress.org/reference/functions/wp_localize_script/? If so, this is not possible, as `wp_localize_script` is working for enqueued/registered assets only. As we don't have a dedicated `.js` file this function cannot be used. This was the only way I found by the current approach. We could extract the JavaScript code still.
2. 👍 
3. 👀 
4. It must be defined in `wp-config.php`. As the translations might be custom and updated from another place than the `core-standards` language files. We could add a fallback or check if the constant is not defined
5. 👍

languages/core-standards-de_DE_formal.po
evemf: I guess this is obsolete since the text has changed and it is done here... https://github.com/netzstrategen/wordpress-core-standards/pull/41. 

@sun or @fabianmarz, am I right?

src/Plugin.php
fabianmarz: I changed the JS to be enqueued instead of rendered inline so `wp_localize_script` works now 👍

andreasblumberg (approved):

PR #40, opened 2019-11-13
Changed WP admin bar to display at bottom of page.

colourgarden:
### Description
- WP admin bar gets fixed to the bottom of the page rather than the top (Copied from Flynt)
- Particularly relevant for WooCommerce notices and Trusted Shops bar where we have had to add custom CSS for handling the Wordpress admin bar. Not ideal
- Also fixed security vulnerability - https://github.com/netzstrategen/wordpress-core-standards/network/alert/yarn.lock/tar/open


sun:
> * WP admin bar gets fixed to the bottom of the page rather than the top (Copied from Flynt)

Why would we want to do that on all sites?

colourgarden:
@sun See point two.

src/UserFrontend.php
```diff
@@ -41,6 +41,9 @@ public static function init() {
     if (!Admin::currentUserHasAccess()) {
       add_filter('show_admin_bar', '__return_false');
     }
+
+    // Moves WP admin bar to bottom of page.
+    add_action('admin_bar_init', __CLASS__ . '::move_admin_bar_to_bottom');
```
LucaPipolo: We are using camel case for custom methods name.

LucaPipolo: Also, why is this into `UserFrontend` and not in `Admin` class?

LucaPipolo (changes requested):
@colourgarden — left a couple of comments.

Also, I get your point about things overlapping on the top of the page but this can also happen at the bottom (e.g. the cookie bar). Did you consider this?

src/UserFrontend.php
colourgarden: All other function names in this file (and project) are using snake case?

The `Admin` class only gets loaded in the admin back-end, not on the front-end (even if the user is an admin). I already tried it.

colourgarden:
@LucaPipolo The high `z-index` value of the admin bar puts it over the top of our cookie notice. I don't think this is something we really have to work around though as a cookie notice usually fixed position and, once dismissed, doesn't return.

The primary reason for moving it to the bottom of the page to avoid a mess with WooCommerce notices and other sticky items at the top of the page. For example - https://github.com/netzstrategen/gaco/blob/master/wp-content/themes/shop/assets/styles/_header.scss#L1-L21

src/UserFrontend.php
LucaPipolo: > All other function names in this file (and project) are using snake case?

Then I do not know why we are not following our standard here. 🤷‍♂ 
Does not matter, just ignore my comment.

> The Admin class only gets loaded in the admin back-end, not on the front-end (even if the user is an admin). I already tried it.

Ah, right. The `Admin` class is hooked on `admin_init`. You're right. 👍

LucaPipolo (approved):

sun:
I don't agree with this change proposal. Moving the admin bar to the bottom makes it harder to control (it always overlaps something, which is not the case at the top), and it introduces a UX inconsistency between backend and frontend.

colourgarden:
@LucaPipolo @sun I've added a constant to the class and a filter to enable this feature to be optional. I don't really agree about the UX concerns as we're only talking about admin editors experiencing the change but I guess it is a bit overreaching to force this on all sites without question.

colourgarden:
OK'd by Luca on the Asana ticket. Will proceed to rebase to squash the commits then merge. Thanks for the feedback.

PR #43, opened 2019-12-18
Fixed incorrect width of cookie consent notice.

evemf:
### Ticket
- [Fix mobile width of overlay](https://app.asana.com/0/873703690981395/1154380372326293/f)


assets/styles/cookie-consent.scss
```diff
@@ -129,12 +129,12 @@
...
 
 .cookie-consent__body {
-  max-width: 60%;
+  max-width: 767px;
   padding: 1em 2em;
   background-color: white;
   color: #666;
 
-  @media screen and (max-width: 550px) {
+  @media screen and (max-width: 990px) {
```
colourgarden: @evemf This doesn't make too much sense to me. Presumably `.cookie-consent__body` is a block-level element so it will get `width: 100%` by default. Then we could just have `max-width: 767px;` at all screen sizes? Then it would take up the full width on smaller devices but we restricted to 767px once the screen is larger than that.

evemf: true, now that you mentioned.... it doesn't make much sense

colourgarden (approved):

PR #50, opened 2020-01-07
Feature Reopen cookie notice

andreasblumberg:
### Ticket
- [Reopen cookie notice](https://app.asana.com/0/809933051638353/1154380372326297/f)

### Description
-


dist/scripts/cookie-consent.js
```diff
@@ -19,9 +25,10 @@
       return;
     }
 
-    var cookie_notice = document.getElementById('cookie-consent');
-    cookie_notice.removeAttribute('hidden');
-    document.body.classList.add('has-cookie-consent');
+    showCookieNotice();
+    var button = document.querySelector('[data-trigger-cookie-consent]');
+    console.log(button);
```
fabianmarz: debug code should be removed 😬 I also wonder why it's added in the dist script only 🤔 maybe added manually?

assets/scripts/cookie-consent.js
```diff
@@ -3,21 +3,34 @@
     return;
   }
 
+  function addCookieSettingsButton() {
```
fabianmarz, line 6: The function name is misleading as this function doesn't actually add a button but registers the click handler for it. IMO this doesn't necessarily needs to be in a dedicated function as it's invoked once. But we can leave it to move forward 👍

dist/scripts/cookie-consent.js
andreasblumberg: fixed

assets/scripts/cookie-consent.js
andreasblumberg: referenced in issue and created a ticket for it to fast forward. #51

fabianmarz (approved):

PR #54, opened 2020-01-27
Added admin email setting as an enviromnent constant.

juanlopez4691:
### Ticket
- [Ensure Wordpress error emails are not sent to client](https://app.asana.com/0/587433704548192/1156992769948077/f)

### Description
-

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -99,6 +99,10 @@ class Plugin {
     add_action('wp_ajax_core-standards/log_cookie_consent', __CLASS__ . '::log_cookie_consent');
     add_action('wp_ajax_nopriv_core-standards/log_cookie_consent', __CLASS__ . '::log_cookie_consent');
 
+    // Sets admin email.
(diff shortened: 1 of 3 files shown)


LucaPipolo (approved):
@juanlopez4691 — code looks good for me but I would add a description in the function comment to explain why we need this feature, what issue we are trying to prevent.

src/Plugin.php
```diff
@@ -271,6 +275,23 @@ public static function log_cookie_consent() {
...
+   * error notices/emails sent from development or staging
+   * environments.
+   *
+   * @param string $value
+   *   Admin email saved in wp_options table.
+   *
+   * @return string
+   *   Admin email.
+   */
+  public static function customAdminEmail($value) {
```
sun, line 291: @juanlopez4691 Would have been nice if this email related adjustment was added to the existing Mail class:

https://github.com/netzstrategen/wordpress-core-standards/blob/master/src/Mail.php

PR #57, opened 2020-02-17
Fixed forgot password form posts to double-prefixed login path.

sun:
### Description
- The form submission redirects to an invalid path (duplicating the custom login path) if the custom login path contains a folder (such as `/my/login`).
- The minor version number is also bumped by this PR, because the PR that introduced the constant forgot to do so.

### DIFF
--- a/src/Security.php
+++ b/src/Security.php
@@ -41,16 +41,16 @@ class Security {
    * @implements wp_redirect
    */
   public static function wp_redirect($url) {
-    $custom_path = defined('CORE_STANDARDS_LOGIN_PATH') ? CORE_STANDARDS_LOGIN_PATH : '/login.php';
+    $custom_path = defined('CORE_STANDARDS_LOGIN_PATH') ? CORE_STANDARDS_LOGIN_PATH : 'login.php';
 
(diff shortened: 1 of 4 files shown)


src/Security.php
```diff
@@ -42,7 +42,16 @@ public static function site_url($url, $path, $scheme) {
    */
   public static function wp_redirect($url) {
     $custom_path = defined('CORE_STANDARDS_LOGIN_PATH') ? CORE_STANDARDS_LOGIN_PATH : '/login.php';
-    $url = strtr($url, ['/wp-login.php' => $custom_path]);
+
+    // wp-login.php calls wp_safe_redirect() with a relative path, which causes
+    // wp_validate_redirect() to automatically prepend the current folder name
+    // of the request URI to it. Prevent the path from being duplicated.
+    if (strpos($url, dirname($custom_path)) === FALSE) {
+      $url = strtr($url, ['wp-login.php' => $custom_path]);
```
fabianmarz, line 50: I removed the leading slash as `wp-login.php` not always add one to the URI https://github.com/WordPress/WordPress/blob/master/wp-login.php#L808

fabianmarz (approved):

PR #60, opened 2020-03-24
Added filename transliteration including non-ASCII characters.

bogdanarizancu:
### Ticket
- [Core-standards: Sanitize/Transliterate uploaded filenames](https://app.asana.com/0/1162879836389371/118154690154506)

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -100,6 +100,9 @@ class Plugin {
     add_filter('pre_option_default_pingback_flag', '__return_zero');
     add_filter('wp_insert_post_data' , __CLASS__ . '::wp_insert_post_data', 100);
 
+     // Transliterate filenames.
+    add_filter('sanitize_file_name', __CLASS__ . '::sanitize_file_name', 10, 2);
+
     if (is_admin()) {
       return;
     }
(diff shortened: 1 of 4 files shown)


src/Plugin.php
```diff
@@ -292,4 +295,13 @@ public static function renderTemplate(array $template_subpathnames, array $varia
     }
   }
 
+  /**
+   * Safari 5 and some mobile browsers do not support umlauts in image URLs yet.
+   *
+   * @implements sanitize_file_name
+   */
+  public static function sanitize_file_name($filename) {
+    return preg_replace('/[^A-Za-z0-9.,_-]+/', '', remove_accents($filename));
```
juanlopez4691, line 304: Do you really need the regex? Isn't `remove_accents()` enough? https://github.com/WordPress/WordPress/blob/7004afe4f4bac1fd17a142051832bdf6be8e6fcf/wp-includes/formatting.php#L1603

bogdanarizancu: I went to [https://gitlab.com/netzstrategen/nland/-/commits/master/wp-content/plugins/custom/src/Plugin.php](https://gitlab.com/netzstrategen/nland/-/commits/master/wp-content/plugins/custom/src/Plugin.php), and saw Sunny's comments, he first used only remove_accents, and then rewrote the function like that and tagged as fixing for covering non-ASCII characters. I can take it out if you think I should, let me know...

juanlopez4691: Ah, OK. Leave as it is then. 👍

juanlopez4691 (approved):

PR #63, opened 2020-05-12
Added git hashes to assets with no version at all.

bogdanarizancu:
### Ticket
- [Performance: Client-side caching for asset and image files (Expires/Cache-Control)](https://app.asana.com/0/809933051638353/912303973823321)

### DIFF
--- a/src/Asset.php
+++ b/src/Asset.php
@@ -30,7 +30,16 @@ class Asset {
     $default_version = $wp_scripts->default_version;
 
     if ($git_version = self::getGitCommitHash()) {
-      $tag = str_replace('?ver=' . $default_version, '?ver=' . $git_version, $tag);
+      $parsed_src = parse_url($src);
+      if (!empty($parsed_src['query'])) {
+        parse_str($parsed_src['query'], $get_args);
+        $get_args['ver'] = $git_version;
+        $new_src = $parsed_src['scheme'] . '://' . $parsed_src['host'] . $parsed_src['path'] . '?' . http_build_query($get_args);
+      }
+      else {
+        $new_src = add_query_arg('ver', $git_version, $src);
+      }
+      $tag = str_replace($src, $new_src, $tag);
     }
     return $tag;
   }
--- a/src/Asset.php
+++ b/src/Asset.php
@@ -30,14 +30,11 @@ class Asset {
     $default_version = $wp_scripts->default_version;
 
     if ($git_version = self::getGitCommitHash()) {
-      $parsed_src = parse_url($src);
-      if (!empty($parsed_src['query'])) {
-        parse_str($parsed_src['query'], $get_args);
-        $get_args['ver'] = $git_version;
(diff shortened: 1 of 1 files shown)


src/Asset.php
```diff
@@ -30,7 +30,16 @@ public static function loader_tag($tag, $handle, $src) {
...
+      $parsed_src = parse_url($src);
+      if (!empty($parsed_src['query'])) {
+        parse_str($parsed_src['query'], $get_args);
+        $get_args['ver'] = $git_version;
+        $new_src = $parsed_src['scheme'] . '://' . $parsed_src['host'] . $parsed_src['path'] . '?' . http_build_query($get_args);
+      }
+      else {
+        $new_src = add_query_arg('ver', $git_version, $src);
+      }
+      $tag = str_replace($src, $new_src, $tag);
```
sun: Parsing and rebuilding the whole URL looks very costly here, also considering the amount of times this is invoked on every page request… did you try something simpler?
```suggestion
      if (strpos($src, '?') !== FALSE) {
        if (preg_match('@[?&](ver=[^&$]*)@', $src, $matches)) {
          $new_src = str_replace($matches[1], 'ver=' . $git_version, $src);
        }
        else {
          $new_src = $src . '&ver=' . $git_version;
        }
      }
      else {
        $new_src = $src . '?ver=' . $git_version;
      }
      $tag = str_replace($src, $new_src, $tag);
```

bogdanarizancu: Well, I used parse_str() and http_build_query() based on our conversation on [slack](https://netzstrategen.slack.com/archives/D0104AWQMED/p1588760326002900). If you think preg_match() + str_replace() are faster than parse_str() + http_build_query(), I won't argue :)

sun: Yeah, I remember that, sounded good in theory until I saw the code :wink: The problem with rebuilding the URL from scratch is that you would have to cater for all possible URLs, e.g. also domain-relative URLs without a hostname. The current code also does not support a port. Or a username/password for basic auth. It is possible to build a URL correctly from the results of parse_str() and take all of these cases into account (see user comments on its man page), but it takes some lines of wonky conditional string concatenation.

The alternative I'm proposing avoids that, but isn't perfect either - it doesn't account for the theoretically possible case of a URL with a #fragment, in which case we wouldn't be able to just append our query string. However, I'm not aware of any real life use-case involving a fragment in asset file URLs (as I don't believe the browser gives the 
loaded asset access to that info), so it's fairly safe to say that this most probably won't happen in practice.

bogdanarizancu: Ok, fair enough, but then I don't see why we would have 2 "else" branches repeating the same line, I compacted them into one with my next fixup. Thanks for taking the time on this one.

sun: The two else branches are different:
```diff
-          $new_src = $src . '&ver=' . $git_version;
+          $new_src = $src . '?ver=' . $git_version;
```

sun: Your code appends a ? even if there is one already.

bogdanarizancu: Bad day today, my mistake, I fixed it.

sun (approved):

PR #64, opened 2020-06-16
Added client-side caching for assets with version query string.

bogdanarizancu:
### Ticket
- [core-standards: Add Expires/Cache-Control header directives to .htaccess for requests with ver= query parameter](https://app.asana.com/0/809933051638353/1176518325514367)


conf/.htaccess.assets.expiry
```diff
@@ -0,0 +1,9 @@
+# BEGIN core-standards:assetsExpiry
+<IfModule mod_expires.c>
```
sun: 1. Is mod_expires available and active on all of our servers?

1. What's the overall distribution / availability between mod_expires and mod_headers?  Would we achieve a better compatibility with mod_headers?

conf/.htaccess.assets.expiry
```diff
@@ -0,0 +1,9 @@
...
+  # Expires + Cache-Control header with a TTL of 1 year for all assets.
+  <FilesMatch "\.(?:css|cur|eot|gif|gz|htc|ico|jpe?g|js|mp4|ogg|ogv|otf|png|svgz?|ttf|webm|woff2?)\?ver">
```
sun: 1. "ver" is not guaranteed to be the only query string parameter, so this regex needs some adjustment. You should also add the terminating = to prevent false-positives on other parameter names.

2. The exclusion for local hostnames is missing? (see issue summary)

src/Schema.php
```diff
@@ -26,6 +26,9 @@ public static function activate() {
 
     // Rename /wp-login.php into /login.php and deny access to XML-RPC API.
     static::ensureFrontControllerAccess();
+
+    // Set assets cache expiry date.
+    static::setAssetsExpiryDate();
```
```suggestion
    // Force a long client-side caching time for assets with "ver" query string.
    static::ensureAssetsCacheControl();
```

src/Schema.php
```diff
@@ -90,6 +93,17 @@ public static function ensureFrontControllerAccess() {
     static::createOrPrependFile(ABSPATH . '.htaccess', $template, "\n");
   }
 
+  /**
+   * Set  Expires + Cache-Control header with a TTL of 1 year for all assets.
+   */
+  public static function setAssetsExpiryDate() {
+    $hostname = parse_url(site_url())['host'];
+    if (strpos($hostname, 'nest') === FALSE && strpos($hostname, 'local') === FALSE && !WP_DEBUG) {
```
sun: The exclusion cannot happen at installation time - it needs to happen at runtime, regardless of where the plugin was installed.

sun (changes requested):

src/Schema.php
sun: Please also use core-standards:assets-cache-control as the delimiter in the .htaccess file and for the template filename.

conf/.htaccess.assets.expiry
bogdanarizancu: Then I guess we can use both, as some sources recommend.

conf/.htaccess.assets.cache.control
```diff
@@ -0,0 +1,10 @@
+# BEGIN core-standards:assets-cache-control
+<IfModule mod_headers.c>
+  # Browser Caching using Cache-Control headers.
+  <If "%{HTTP_HOST} !~ /local|nest/">
+    <FilesMatch "\.(?:css|cur|eot|gif|gz|htc|ico|jpe?g|js|mp4|ogg|ogv|otf|png|svgz?|ttf|webm|woff2?).*[?&]ver=">
+      Header set Cache-Control "max-age=31536000, public"
```
sun: The cacheability is normally specified first, because it can make all other parameters obsolete
```suggestion
      Header set Cache-Control "public, max-age=31536000"
```
See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

conf/.htaccess.assets.cache.control
```diff
@@ -0,0 +1,10 @@
+# BEGIN core-standards:assets-cache-control
+<IfModule mod_headers.c>
+  # Browser Caching using Cache-Control headers.
```
sun:
```suggestion
  # Client-side caching for assets with version query string.
  # Not applied locally as Git hash does not change while editing.
```

conf/.htaccess.assets.cache.control
```diff
@@ -0,0 +1,10 @@
...
+  <If "%{HTTP_HOST} !~ /local|nest/">
```
sun: Missing dot and end of string anchor
```suggestion
  <If "%{HTTP_HOST} !~ /\.(local|nest)$/">
```

src/Schema.php
```diff
@@ -90,6 +93,14 @@ public static function ensureFrontControllerAccess() {
     static::createOrPrependFile(ABSPATH . '.htaccess', $template, "\n");
   }
 
+  /**
+   * Set Cache-Control header with a TTL of 1 year for all assets.
```
sun: It's perfectly okay to repeat the inline comment from init as PHPDoc summary line if there's nothing further to say.
```suggestion
   * Force a long client-side caching time for assets with "ver" query string.
```

src/Schema.php
```diff
@@ -90,6 +93,14 @@ public static function ensureFrontControllerAccess() {
...
+   */
+  public static function ensureAssetsCacheControl() {
+    $template = file_get_contents(Plugin::getBasePath() . '/conf/.htaccess.assets.cache.control');
```
sun: Can we remove the "control"? It's too specific.  Sorry for the change of mind 🙈 
```suggestion
    $template = file_get_contents(Plugin::getBasePath() . '/conf/.htaccess.assets.cache');
```
The block delimiter can change accordingly into `core-standards:assets-cache`

sun (changes requested):

conf/.htaccess.assets.cache
```diff
@@ -0,0 +1,11 @@
+# BEGIN core-standards:assets-cache
+<IfModule mod_headers.c>
+  #Client-side caching for assets with version query string.
```
sun:
```suggestion
  # Client-side caching for assets with version query string.
```

sun (approved):
Thanks, looks good to me!

Can we test this on one of our staging sites before merging? (update submodule, deactivate plugin, activate plugin)

core-standards.php
```diff
@@ -2,7 +2,7 @@
 
 /*
   Plugin Name: Core Standards
-  Version: 2.3.3
+  Version: 2.4
```
sun: https://semver.org/
```suggestion
  Version: 2.4.0
```

conf/.htaccess.assets.cache
```diff
@@ -0,0 +1,13 @@
+# BEGIN core-standards:assets-cache
+<IfModule mod_headers.c>
+  # Client-side caching for assets with version query string.
+  # Not applied locally as Git hash does not change while editing.
+  <FilesMatch "\.(?:css|cur|eot|gif|gz|htc|ico|jpe?g|js|mp4|ogg|ogv|otf|png|svgz?|ttf|webm|woff2?)">
+    <If "%{HTTP_HOST} =~ /(local|nest)/">
+    </If>
+    <ElseIf "%{QUERY_STRING} =~ /ver=/">
+      Header set Cache-Control "public, max-age=31536000"
+    </ElseIf>
```
sun: Nested conditions should work, no?

Also added the delimiter ?& before ver.
```suggestion
    <If "%{HTTP_HOST} =~ /(local|nest)/">
      <If "%{QUERY_STRING} =~ /[?&]ver=/">
        Header set Cache-Control "public, max-age=31536000"
      </If>
    </If>
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "core-standards",
-  "version": "2.3.3",
+  "version": "2.4",
```
sun:
```suggestion
  "version": "2.4.0",
```

README.txt
```diff
@@ -25,6 +25,8 @@ Core already but are not for different reasons (as the name implies).
   virtual host or the whole server (the latter is not recommended for production
   servers).
 
+  Adds long client-side caching time for assets with "ver" query string via .htacces.
```
sun:
```suggestion
- Adds client-side caching for assets having a `'ver'` query string.
```

sun (changes requested):

conf/.htaccess.assets.cache
bogdanarizancu: - I've been testing all possible combinations on `netz-stage` and, as commented in Asana, nested <If> directives do not work, this is why I used this <ElseIf> workaround, which does work even if it is not the prettiest

- as for `QUERY_STRING`, the usual case is something `ver=d5e4f753`, which would not match `/[?&]ver=/` (I tested it).

sun: Quickly tested on https://stage.netz.netz.rocks/, this works:
```suggestion
    <If "%{HTTP_HOST} !~ /\.(local|nest)$/ && %{QUERY_STRING} =~ /(^|&)ver=/">
      Header set Cache-Control "public, max-age=31536000"
    </If>
```
If expressions support logical AND and OR conditions. See https://httpd.apache.org/docs/trunk/expr.html

Also fixed the comparison operator for the host name.

Of course, QUERY_STRING only contains the query string itself without its delimiter from the URL, so the ? is not part of it.  It is replaced with the start of string anchor ^.  However, that is not an actual character, so the character range [] cannot be used and needs to be replaced with multiple subgroups (|).

You can temporarily add `|rocks` to confirm that the domain exclusion works.

sun (approved):

PR #65, opened 2020-06-19
Fixed PHP files in wp-includes can be executed, existing .htaccess snippets are not updated.

juanlopez4691:
### Ticket
- [Fatal errors are logged in production](https://app.asana.com/0/1156709897787098/1180108570410600)

### Description
-

### DIFF
--- a/conf/.htaccess.security
+++ b/conf/.htaccess.security
@@ -1,4 +1,6 @@
 # BEGIN core-standards:security
+RedirectMatch 403 /wp-includes/.+\.php$
+
 <IfModule mod_rewrite.c>
   # Rename /wp-login.php into /login.php and deny access to XML-RPC API.
   RewriteEngine On
--- a/src/Schema.php
+++ b/src/Schema.php
@@ -121,13 +121,13 @@ class Schema {
     else {
       $content = file_get_contents($pathname);
       if (FALSE === strpos($content, $template)) {
-        // Remove old content if present, looking for the template block start
-        // and end lines.
+        // Remove old content if present, looking for the template block's
+        // BEGIN and END lines.
         $templateLines = explode("\n", rtrim($template, "\n"));
         $start = strpos($content, $templateLines[0]);
         $end = strpos($content, end($templateLines));
         if ($start !== FALSE && $end !== FALSE) {
-          $content = substr_replace($content, '', $start, strlen($template));
+          $content = substr_replace($content, '', $start, $end - $start + 1);
         }
 
         $content = $template . $separator . $content;
--- a/src/Schema.php
+++ b/src/Schema.php
@@ -123,14 +123,15 @@ class Schema {
       if (FALSE === strpos($content, $template)) {
         // Remove old content if present, looking for the template block's
         // BEGIN and END lines.
-        $templateLines = explode("\n", rtrim($template, "\n"));
-        $start = strpos($content, $templateLines[0]);
-        $end = strpos($content, end($templateLines));
-        if ($start !== FALSE && $end !== FALSE) {
-          $content = substr_replace($content, '', $start, $end - $start + 1);
+        $template_lines = explode("\n", rtrim($template, "\n"));
+        $begin = strpos($content, $template_lines[0]);
+        $end = strpos($content, end($template_lines));
+        if ($begin !== FALSE && $end !== FALSE) {
+          $content = substr_replace($content, $template, $begin, $end - $begin + 1);
         }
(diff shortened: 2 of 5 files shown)


conf/.htaccess.security
```diff
@@ -1,4 +1,6 @@
 # BEGIN core-standards:security
+RedirectMatch 403 /wp-includes/.+\.php$
```
sun, line 2: To my knowledge, we do not have a routine for updating these snippets yet, so just updating the template will not update the site. We added the begin/end delimiters to enable that. Could you look into that?

sun (changes requested):

src/Schema.php
```diff
@@ -121,6 +121,15 @@ private static function createOrPrependFile($pathname, $template, $separator = '
     else {
       $content = file_get_contents($pathname);
       if (FALSE === strpos($content, $template)) {
+        // Remove old content if present, looking for the template block start
+        // and end lines.
```
```suggestion
        // Remove old content if present, looking for the template block's
        // BEGIN and END lines.
```

src/Schema.php
```diff
@@ -121,6 +121,15 @@ private static function createOrPrependFile($pathname, $template, $separator = '
...
+        $templateLines = explode("\n", rtrim($template, "\n"));
+        $start = strpos($content, $templateLines[0]);
+        $end = strpos($content, end($templateLines));
+        if ($start !== FALSE && $end !== FALSE) {
+          $content = substr_replace($content, '', $start, strlen($template));
```
sun: I would have expected to see something in relation to `$end` here in order to remove only the amount of lines of the old/previously existing snippet…?

juanlopez4691:
@sun We will have to fix the existing template file `.htaccess.uploads.noscript` so that it includes the start/end block marks as expected (`# BEGIN core-standards:uploads.noscript`, `# END core-standards:uploads.noscript`). Those are not included in the current version of the plugin. This fix can be part of the `core-standards` update in each site.

src/Schema.php
```diff
@@ -121,6 +121,15 @@ private static function createOrPrependFile($pathname, $template, $separator = '
...
+        // BEGIN and END lines.
+        $templateLines = explode("\n", rtrim($template, "\n"));
+        $start = strpos($content, $templateLines[0]);
+        $end = strpos($content, end($templateLines));
+        if ($start !== FALSE && $end !== FALSE) {
+          $content = substr_replace($content, '', $start, $end - $start + 1);
+        }
+
         $content = $template . $separator . $content;
         $write_content = TRUE;
```
sun: 1. Let's update the existing snippet instead of removing it and inserting the new one at a different position.
1. Let's be consistent with begin/end (and don't mix start into it).
1. Local variable names should be in snake_case.
```suggestion
        // Remove old content if present, looking for the template block's
        // BEGIN and END lines.
        $template_lines = explode("\n", rtrim($template, "\n"));
        $begin = strpos($content, $template_lines[0]);
        $end = strpos($content, end($template_lines));
        if ($begin !== FALSE && $end !== FALSE) {
          $content = substr_replace($content, $template, $begin, $end - $begin + 1);
        }
        else {
          $content = $template . $separator . $content;
        }
        $write_content = TRUE;
```

juanlopez4691:
@sun I was not considering the length of the last line of the block to be injected/substituted. This is now fixed.

Tested locally and in stage.gaco, works nicely.

juanlopez4691:
@sun Can you please review last changes? It would be nice to close this 🙏 😄

src/Schema.php
```diff
@@ -121,7 +121,18 @@ private static function createOrPrependFile($pathname, $template, $separator = '
...
       if (FALSE === strpos($content, $template)) {
-        $content = $template . $separator . $content;
+        // Remove old content if present, looking for the template block's
+        // BEGIN and END lines.
+        $template_lines = explode("\n", rtrim($template, "\n"));
+        $begin = strpos($content, $template_lines[0]);
+        $lastLine = end($template_lines);
+        $end = strpos($content, $lastLine) + strlen($lastLine);
+        if ($begin !== FALSE && $end !== FALSE) {
+          $content = substr_replace($content, $template, $begin, $end - $begin + 1);
```
sun, line 131: I still don't really understand how this works; I would simply do it with a preg_replace() on the $content instead… but I trust that you tested it extensively (and because you understand it, you can maintain it in the future 😀)

sun (approved):
I adjusted the PR title.

PR #67, opened 2020-08-21
Added security directive to disallow access to .git folders.

bogdanarizancu:
### Ticket
- [Update core-standards to add this in .htaccess as a security measure](https://app.asana.com/0/809933051638353/1190062271672034)

### DIFF
--- a/conf/.htaccess.security
+++ b/conf/.htaccess.security
@@ -7,5 +7,8 @@ RedirectMatch 403 /wp-includes/.+\.php$
   RewriteRule ^login\.php$ /wp-login.php [QSA,END]
   RewriteRule ^wp-login\.php$ - [NS,F,END]
   RewriteRule ^xmlrpc\.php$ - [F,END]
+
+  #Ensure .git folder is not accessible
+  RewriteRule \.git/.*?$ - [F,NC]
 </IfModule>
 # END core-standards:security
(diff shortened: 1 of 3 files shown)


conf/.htaccess.security
```diff
@@ -7,5 +7,8 @@ RedirectMatch 403 /wp-includes/.+\.php$
   RewriteRule ^login\.php$ /wp-login.php [QSA,END]
   RewriteRule ^wp-login\.php$ - [NS,F,END]
   RewriteRule ^xmlrpc\.php$ - [F,END]
+
+  #Ensure .git folder is not accessible
+  RewriteRule \.git/.*?$ - [F,NC]
```
sun: I don't understand what you're matching after the directory.
```suggestion
  RewriteRule \.git/ - [F,END]
```
Note that this also applies (and should also apply) to other .git folders in the filesystem; e.g., `wp-content/plugins/core-standards/.git/HEAD`

conf/.htaccess.security
```diff
@@ -7,5 +7,8 @@ RedirectMatch 403 /wp-includes/.+\.php$
   RewriteRule ^login\.php$ /wp-login.php [QSA,END]
   RewriteRule ^wp-login\.php$ - [NS,F,END]
   RewriteRule ^xmlrpc\.php$ - [F,END]
+
+  #Ensure .git folder is not accessible
```
sun:
```suggestion
  # Disallow access to .git folders.
```
This should also be moved to the leading comment of this block.

sun (approved):

PR #68, opened 2020-09-16
Fixed version hash was added to remote assets.

juanlopez4691:
### Ticket
- [Fix after-update issue: paypal button is missing on cart page](https://app.asana.com/0/1156709897787098/1187794656912659)

### Description
- Related to https://github.com/netzstrategen/gaco/pull/539


bogdanarizancu (approved):
Thanks for this update, definitely much needed 👍 
One question would be if we don't usually check `strpos` for `FALSE` instead of `0`? 
This as approved whatever the case.

PR #69, opened 2020-11-11
Fixed WordPress admin UI is cluttered with messages notifying about new releases and more features.

Mauricio-Urrego:
### Ticket
- [Remove WordPress release update and WooCommerce.com bragging notices](https://app.asana.com/0/354170587449545/1197780930151545/f)

### Description
-

### DIFF
--- a/src/Admin.php
+++ b/src/Admin.php
@@ -72,9 +72,9 @@ class Admin {
    * @see http://premium.wpmudev.org/blog/hide-the-wordpress-update-notification/
    */
   public static function removeUselessCoreUpdateNagMessages() {
-      // @see wp-admin/includes/admin-filters.php
-      remove_action('admin_notices', 'update_nag', 3);
-      remove_action('admin_notices', 'maintenance_nag', 10);
+    // @see wp-admin/includes/admin-filters.php
+    remove_action('admin_notices', 'update_nag', 3);
+    remove_action('admin_notices', 'maintenance_nag', 10);
   }
 
   /**
--- a/src/Admin.php
+++ b/src/Admin.php
@@ -72,11 +72,9 @@ class Admin {
    * @see http://premium.wpmudev.org/blog/hide-the-wordpress-update-notification/
    */
   public static function removeUselessCoreUpdateNagMessages() {
-    if (!current_user_can('update_core')) {
       // @see wp-admin/includes/admin-filters.php
       remove_action('admin_notices', 'update_nag', 3);
       remove_action('admin_notices', 'maintenance_nag', 10);
-    }
   }
 
   /**


Mauricio-Urrego:
Working on removing woocommerce notices now.

src/Admin.php
```diff
@@ -71,12 +71,12 @@ public static function init() {
...
    */
-  public static function removeUselessCoreUpdateNagMessages() {
-    if (!current_user_can('update_core')) {
+  public static function removeNagNotices() {
       // @see wp-admin/includes/admin-filters.php
       remove_action('admin_notices', 'update_nag', 3);
       remove_action('admin_notices', 'maintenance_nag', 10);
-    }
+      //removes all nag notices including from woocommerce
+      remove_all_actions('admin_notices');
```
sun: We cannot remove all admin notices, because this would remove any kind of feedback for the user's operations, including error messages and success/completion messages.

src/Admin.php
```diff
@@ -71,12 +71,12 @@ public static function init() {
    * @see http://www.paulund.co.uk/hide-wordpress-update-notice
    * @see http://premium.wpmudev.org/blog/hide-the-wordpress-update-notification/
    */
-  public static function removeUselessCoreUpdateNagMessages() {
-    if (!current_user_can('update_core')) {
+  public static function removeNagNotices() {
       // @see wp-admin/includes/admin-filters.php
       remove_action('admin_notices', 'update_nag', 3);
       remove_action('admin_notices', 'maintenance_nag', 10);
```
sun: The indentation of the code needs to updated accordingly.

src/Admin.php
```diff
@@ -71,12 +71,12 @@ public static function init() {
...
-    }
+      //removes all nag notices including from woocommerce
```
sun: All code comments (as well as commit messages) should form proper English sentences, starting with an capitalized letter and ending with punctuation.  Have a look at the rest of the file to see many good examples.

We also have (or had) these guidelines written down somewhere, which would be a better read than getting one by one feedback about individual coding style issues in your PRs, but I'm not able to find locate them right now (see Slack).

sun (changes requested):
Unfortunately that approach is a bit too greedy 😅   I would recommend to use one of the options mentioned in the Asana task instead.

There are also various coding style issues.  We always clean up the code before asking others for peer reviews - that is, unless a PR is an early draft or just a throw-away-prototype to get a basic idea across.  As long as we're trying to find our existing documentation about coding style, the most straightforward advice I can give you is to simply orient yourself at the surrounding lines of the same file.  Because within a single file, the applied coding standard should be consistent.  So if you follow the rules of existing indentations, comments, whitespace, and line wrapping of the other lines in the file, then your code should be 90% safe already.

I also revised the PR title to state more clearly what the change is about. All commits in the branch will be squashed and merged into the mainline using the PR title as commit message in the end.

src/Admin.php
```diff
@@ -72,11 +72,9 @@ public static function init() {
    * @see http://premium.wpmudev.org/blog/hide-the-wordpress-update-notification/
    */
   public static function removeUselessCoreUpdateNagMessages() {
-    if (!current_user_can('update_core')) {
       // @see wp-admin/includes/admin-filters.php
       remove_action('admin_notices', 'update_nag', 3);
       remove_action('admin_notices', 'maintenance_nag', 10);
-    }
   }
```
sun: @Mauricio-Urrego The indentation of the resulting code is still not right - the indentation level is two spaces and the code should not diverge from that:

image

sun (changes requested):

sun (approved):

PR #71, opened 2021-12-07
Added security headers to all responses.

chrisbaltazar:
### Ticket
https://app.asana.com/0/30156186242982/1201478743538123/f)

### Description
Added some additional security headers to be compliant with the latest regulations.

### DIFF
--- a/conf/.htaccess.headers
+++ b/conf/.htaccess.headers
@@ -2,8 +2,8 @@
 <IfModule mod_headers.c>
   Header set Content-Security-Policy "frame-ancestors"
   Header set Referrer-Policy "no-referrer-when-downgrade"
+  Header set Strict-Transport-Security "Strict-Transport-Security: max-age=31536000; includeSubDomains"
   Header set X-Content-Type-Options "nosniff"
   Header set X-XSS-Protection "1; mode=block"
-  Header set Strict-Transport-Security "Strict-Transport-Security: max-age=31536000; includeSubDomains"
 </IfModule>
 # END core-standards:headers
--- a/conf/.htaccess.headers
+++ b/conf/.htaccess.headers
@@ -1,9 +1,9 @@
 # BEGIN core-standards:headers
 <IfModule mod_headers.c>
- Header set Content-Security-Policy "frame-ancestors"
- Header set Referrer-Policy "no-referrer-when-downgrade"
- Header set X-Content-Type-Options "nosniff"
- Header set X-XSS-Protection "1; mode=block"
- Header set Strict-Transport-Security "Strict-Transport-Security: max-age=31536000; includeSubDomains"
+  Header set Content-Security-Policy "frame-ancestors"
+  Header set Referrer-Policy "no-referrer-when-downgrade"
+  Header set X-Content-Type-Options "nosniff"
+  Header set X-XSS-Protection "1; mode=block"
+  Header set Strict-Transport-Security "Strict-Transport-Security: max-age=31536000; includeSubDomains"
 </IfModule>
 # END core-standards:headers
--- a/conf/.htaccess.headers
+++ b/conf/.htaccess.headers
@@ -1,0 +1,9 @@
+# BEGIN core-standards:headers
+<IfModule mod_headers.c>
+ Header set Content-Security-Policy "frame-ancestors"
+ Header set Referrer-Policy "no-referrer-when-downgrade"
+ Header set X-Content-Type-Options "nosniff"
+ Header set X-XSS-Protection "1; mode=block"
+ Header set Strict-Transport-Security "Strict-Transport-Security: max-age=31536000; includeSubDomains"
+</IfModule>
+# END core-standards:headers
--- a/conf/.htaccess.headers
+++ b/conf/.htaccess.headers
@@ -1,9 +1,0 @@
-# BEGIN core-standards:headers
-<IfModule mod_headers.c>
-  Header set Content-Security-Policy "frame-ancestors"
-  Header set Referrer-Policy "no-referrer-when-downgrade"
-  Header set Strict-Transport-Security "Strict-Transport-Security: max-age=31536000; includeSubDomains"
-  Header set X-Content-Type-Options "nosniff"
-  Header set X-XSS-Protection "1; mode=block"
-</IfModule>
-# END core-standards:headers
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -48,7 +48,6 @@ class Plugin {
     add_filter('network_site_url', __NAMESPACE__ . '\Security::site_url', 100, 3);
     add_filter('site_url', __NAMESPACE__ . '\Security::site_url', 100, 3);
     add_filter('wp_redirect', __NAMESPACE__ . '\Security::wp_redirect', 1);
-    add_filter('wp_headers', __NAMESPACE__ . '\Security::additional_securityheaders', 1);
 
     Plugin::loadTextdomain();
   }
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -48,6 +48,7 @@ class Plugin {
     add_filter('network_site_url', __NAMESPACE__ . '\Security::site_url', 100, 3);
     add_filter('site_url', __NAMESPACE__ . '\Security::site_url', 100, 3);
     add_filter('wp_redirect', __NAMESPACE__ . '\Security::wp_redirect', 1);
+    add_filter('wp_headers', __NAMESPACE__ . '\Security::additional_securityheaders', 1);
 
     Plugin::loadTextdomain();
   }
--- a/src/Schema.php
+++ b/src/Schema.php
@@ -75,13 +75,9 @@ class Schema {
   /**
    * Performs a preg match to replace the new header value in the stream.
    */
-  private static function findAndReplaceHeader(string &$content, string $headerName, string $headerValue) {
-    $search = sprintf('<(%s) \"(.*)\">', $headerName);
-    // Replace using the same header name and value.
-    $replacement = sprintf('$1 "%s"', $headerValue);
-    $result = preg_replace($search, $replacement, $content);
+  private static function findAndReplaceHeader(string &$content, string $header_name, string $header_value) {
     // Ensures there was no error in the matching otherwise keeps original.
-    $content = $result ?? $content;
+    $content = preg_replace("@($header_name) .*@", "$1 $header_value", $content) ?? $content;
   }
 
   /**
--- a/src/Schema.php
+++ b/src/Schema.php
@@ -60,14 +60,14 @@ class Schema {
     if (defined('CORE_STANDARDS_REFERRER_POLICY')) {
       static::findAndReplaceHeader($template, 'Referrer-Policy', CORE_STANDARDS_REFERRER_POLICY);
     }
+    if (defined('CORE_STANDARDS_STRICT_TRANSPORT_SECURITY')) {
+      static::findAndReplaceHeader($template, 'Strict-Transport-Security', CORE_STANDARDS_STRICT_TRANSPORT_SECURITY);
+    }
     if (defined('CORE_STANDARDS_X_CONTENT_TYPE_OPTIONS')) {
       static::findAndReplaceHeader($template, 'X-Content-Type-Options', CORE_STANDARDS_X_CONTENT_TYPE_OPTIONS);
     }
     if (defined('CORE_STANDARDS_X_XSS_PROTECTION')) {
       static::findAndReplaceHeader($template, 'X-XSS-Protection', CORE_STANDARDS_X_XSS_PROTECTION);
-    }
-    if (defined('CORE_STANDARDS_STRICT_TRANSPORT_SECURITY')) {
-      static::findAndReplaceHeader($template, 'Strict-Transport-Security', CORE_STANDARDS_STRICT_TRANSPORT_SECURITY);
     }
     static::createOrPrependFile(ABSPATH . '.htaccess', $template, "\n");
   }
--- a/src/Schema.php
+++ b/src/Schema.php
@@ -13,10 +13,11 @@ namespace Netzstrategen\CoreStandards;
 class Schema {
 
   /**
-   * Default http headers useful to prevent clickjacking issues.
-   * Please note how the value of each header must include the quotes when needed.
-   * In case any specific project wants to extend/override these from the config/env
-   * Using the constant called CORE_STANDARDS_HTTP_HEADERS.
+   * HTTP headers to prevent clickjacking, XSS, and other vulnerabilities.
+   *
+   * The value for each header must include quotes where needed.
+   * The defaults can be overridden by defining a constant named
+   * CORE_STANDARDS_HTTP_HEADERS.
    */
   const HTTP_RESPONSE_HEADERS = [
     'Content-Security-Policy' => '"frame-ancestors"',
--- a/src/Schema.php
+++ b/src/Schema.php
@@ -72,13 +72,13 @@ class Schema {
     if (defined('CORE_STANDARDS_HTTP_HEADERS') && is_array(CORE_STANDARDS_HTTP_HEADERS)) {
       $headers = array_merge($headers, CORE_STANDARDS_HTTP_HEADERS);
     }
-    $template = '# BEGIN core-standards:headers' . PHP_EOL;
-    $template .= '<IfModule mod_headers.c>' . PHP_EOL;
+    $template = '# BEGIN core-standards:headers\n';
+    $template .= '<IfModule mod_headers.c>\n';
     foreach ($headers as $header => $value) {
-      $template .= "\tHeader set $header $value" . PHP_EOL;
+      $template .= "\tHeader set $header $value\n";
     }
-    $template .= '</IfModule>' . PHP_EOL;
-    $template .= '# END core-standards:headers' . PHP_EOL;
+    $template .= '</IfModule>\n';
+    $template .= '# END core-standards:headers\n';
     static::createOrPrependFile(ABSPATH . '.htaccess', $template, "\n");
   }
 
--- a/src/Schema.php
+++ b/src/Schema.php
@@ -29,6 +29,9 @@ class Schema {
 
     // Force a long client-side caching time for assets with "ver" query string.
     static::ensureAssetsCacheControl();
+
+    // Includes the necessary response headers for all requests in htaccess file.
+    static::ensureResponseHeadersHtaccess();
   }
 
   /**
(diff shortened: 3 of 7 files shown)


src/Security.php
```diff
@@ -55,4 +55,21 @@ public static function wp_redirect($url) {
...
   }
 
+  /**
+   * @implemements wp_headers
+   *
+   * @return array
+   */
+  public static function additional_securityheaders(array $headers):array {
+    if (!is_admin()) {
+      $headers['Referrer-Policy']           ??= defined('CORE_STANDARDS_Referrer_Policy') ? CORE_STANDARDS_Referrer_Policy : 'no-referrer-when-downgrade';
```
sun: 1. Usage of the [null coalesce assignment operator](https://www.php.net/manual/en/language.operators.assignment.php#language.operators.assignment.other) is wrong here; the tested value must be variable that may be null. `defined()` is a language expression and never null but true or false.

2. PHP constant names must always be all-uppercase.

3. Can you sort the assignments alphabetically by header name, please?  It's not clear in which order they are currently.

4. Minor: The equal sign of the assignments on multiple lines should not be aligned. It causes an unnecessary line changes in future PRs.

src/Security.php
```diff
@@ -55,4 +55,21 @@ public static function wp_redirect($url) {
     return $url;
   }
 
+  /**
+   * @implemements wp_headers
+   *
+   * @return array
+   */
+  public static function additional_securityheaders(array $headers):array {
```
sun: 1. Minor: Missing space before return type hint.
    ```suggestion
      public static function additional_securityheaders(array $headers): array {
    ```

2. Minor: Missing blank line after function declaration (before closing class).

src/Security.php
```diff
@@ -55,4 +55,21 @@ public static function wp_redirect($url) {
...
+    if (!is_admin()) {
+      $headers['Referrer-Policy']           ??= defined('CORE_STANDARDS_Referrer_Policy') ? CORE_STANDARDS_Referrer_Policy : 'no-referrer-when-downgrade';
+      $headers['X-Content-Type-Options']    ??= defined('CORE_STANDARDS_X_Content_Type_Options') ? CORE_STANDARDS_X_Content_Type_Options : 'nosniff';
+      $headers['X-XSS-Protection']          ??= defined('CORE_STANDARDS_X_XSS_Protection') ? CORE_STANDARDS_X_XSS_Protection : '1; mode=block';
+      $headers['Permissions-Policy']        ??= defined('CORE_STANDARDS_Permissions_Policy') ? CORE_STANDARDS_Permissions_Policy : 'geolocation=(self "https://example.com") microphone=() camera=()';
+      $headers['Content-Security-Policy']   ??= defined('CORE_STANDARDS_Content_Security_Policy') ? CORE_STANDARDS_Content_Security_Policy : 'frame-ancestors';
```
sun: It is my understanding that the Content Security Policy header replaces most of the other headers, such as X-Frame-Options. Shouldn't we skip the legacy headers and directly implement all directives in the CSP instead?

src/Security.php
```diff
@@ -55,4 +55,21 @@ public static function wp_redirect($url) {
...
+   * @implemements wp_headers
+   *
+   * @return array
+   */
+  public static function additional_securityheaders(array $headers):array {
+    if (!is_admin()) {
+      $headers['Referrer-Policy']           ??= defined('CORE_STANDARDS_Referrer_Policy') ? CORE_STANDARDS_Referrer_Policy : 'no-referrer-when-downgrade';
+      $headers['X-Content-Type-Options']    ??= defined('CORE_STANDARDS_X_Content_Type_Options') ? CORE_STANDARDS_X_Content_Type_Options : 'nosniff';
+      $headers['X-XSS-Protection']          ??= defined('CORE_STANDARDS_X_XSS_Protection') ? CORE_STANDARDS_X_XSS_Protection : '1; mode=block';
+      $headers['Permissions-Policy']        ??= defined('CORE_STANDARDS_Permissions_Policy') ? CORE_STANDARDS_Permissions_Policy : 'geolocation=(self "https://example.com") microphone=() camera=()';
```
sun: Should use the site URL instead of example.com.
```suggestion
      $site_url = site_url();
      $headers['Permissions-Policy']        ??= defined('CORE_STANDARDS_Permissions_Policy') ? CORE_STANDARDS_Permissions_Policy : 'geolocation=(self "' . $site_url . '") microphone=() camera=()';
```

src/Plugin.php
```diff
@@ -48,6 +48,7 @@ public static function plugins_loaded() {
     add_filter('network_site_url', __NAMESPACE__ . '\Security::site_url', 100, 3);
     add_filter('site_url', __NAMESPACE__ . '\Security::site_url', 100, 3);
     add_filter('wp_redirect', __NAMESPACE__ . '\Security::wp_redirect', 1);
+    add_filter('wp_headers', __NAMESPACE__ . '\Security::additional_securityheaders', 1);
```
sun: 1. The Content Security Policy is a different kind of security layer then existing stuff here (which are protections against script kiddies).  Can you move this into a new ContentSecurity class?

2. The callback method name should match the hook name, or it should otherwise clarify what is doing (adding response headers).

3. Why did you use the priority 1?  Whenever using a priority different to the default, there should be an inline comment that explains before or after which other thing this registered hook is supposed to run.  Other code can change their priorities, at which point it becomes hard to debug what the originally intended integration was.  Given that there is no comment or commit stating something in relation to it, I assume the priority 1 was a premature optimization, so I would recommend to remove it.

```suggestion
    add_filter('wp_headers', __NAMESPACE__ . '\ContentSecurity::wp_headers');
```

sun (changes requested):
As mentioned on the Asana task, the core-standards plugin ships with a framework for adding content blocks to the .htaccess file.  If we want this to work for all responses and not just PHP responses (which you seem to agree with on the Asana task), then we need to migrate this code to integrate with the .htaccess framework:
https://github.com/netzstrategen/wordpress-core-standards/blob/master/src/Schema.php#L21-L31

The values would still need to be dynamic.  Therefore, a template file as in the other cases might be unnecessary.  Or perhaps we can add one and only replace placeholder variables in it with dynamic values?

src/Security.php
chrisbaltazar: Agree in all your points except the first one, the null coalesce operator is totally fine here, the evaluated value is the one coming from the ARRAY in the first place, so the intention is to respect any previous value set when arriving to this point, here the proof https://3v4l.org/teZs4

src/Security.php
chrisbaltazar: sorry my bad

src/Security.php
chrisbaltazar: I can not find any documentation about what you state here, but I trust you if you want me to do it like that, could you please just tell which other headers are replaced by this one?

src/Plugin.php
chrisbaltazar: In this case I've changed the original approach I took and followed your suggestion to use the htaccess rules instead, so I moved all the logic from here to the existing `Schema` class which is the one in charge of this job.

src/Security.php
chrisbaltazar: I've removed this header for now giving the nature of it I think it may cause some other unwanted issues.

src/Schema.php
```diff
@@ -46,6 +49,41 @@ public static function uninstall() {
...
+      static::findAndReplaceHeader($template, 'X-XSS-Protection', CORE_STANDARDS_X_XSS_PROTECTION);
+    }
+    static::createOrPrependFile(ABSPATH . '.htaccess', $template, "\n");
+  }
+
+  /**
+   * Performs a preg match to replace the new header value in the stream.
+   */
+  private static function findAndReplaceHeader(string &$content, string $headerName, string $headerValue) {
+    $search = sprintf('<(%s) \"(.*)\">', $headerName);
```
sun: Potentially existing PCRE control chars should be quoted.
```suggestion
    $search = sprintf('<(%s) \"(.*)\">', preg_quote($headerName, '/'));
```
Usually we're using the `@` sign as pattern delimiters; the delimiters `<>` are seldomly used in the wild.
```suggestion
    $search = sprintf('@(%s) \"(.*)\"@', preg_quote($headerName, '/'));
```

src/Schema.php
```diff
@@ -46,6 +49,41 @@ public static function uninstall() {
...
+    // Replace using the same header name and value.
+    $replacement = sprintf('$1 "%s"', $headerValue);
+    $result = preg_replace($search, $replacement, $content);
```
sun: Maybe it is just me, but somehow I find the additional separation into `sprintf()` expressions less readable than a direct preg_replace()…?
```suggestion
    $result = preg_replace("@($headerName) \"(.*)\"@", "$1 \"$headerValue\"", $content);
```

Also note that the expression can be simplified: https://regex101.com/r/jQNTde/1 - in particular, the dot does not match a newline unless the modifier "s" is passed.
```suggestion
    $result = preg_replace("@($headerName) .*@", "$1 \"$headerValue\"", $content);
```

Also note that we need to be able to add header expressions/conditions after the value. You can see an example in the headers mentioned in the directives that were referenced in the task's issue summary:
https://github.com/netzstrategen/nutr-nmed/blob/master/.htaccess#L193

This means the quotes should be part of the value that is in the constant:
```suggestion
    $headerName = preg_quote($headerName);
    $result = preg_replace("@($headerName) .*@", "$1 $headerValue", $content);
```
In this final form, the direct preg_replace() definitely looks much easier to read than the separate lines.

src/Schema.php
```diff
@@ -46,6 +49,41 @@ public static function uninstall() {
...
+    if (defined('CORE_STANDARDS_X_XSS_PROTECTION')) {
+      static::findAndReplaceHeader($template, 'X-XSS-Protection', CORE_STANDARDS_X_XSS_PROTECTION);
+    }
+    static::createOrPrependFile(ABSPATH . '.htaccess', $template, "\n");
+  }
+
+  /**
+   * Performs a preg match to replace the new header value in the stream.
+   */
+  private static function findAndReplaceHeader(string &$content, string $headerName, string $headerValue) {
```
sun: Minor: Local variable names within functions should be in `$snake_case` (as seen in the rest of the file).  Only class properties and methods use lowerCamelCase.
```suggestion
  private static function findAndReplaceHeader(string &$content, string $header_name, string $header_value) {
```

src/Schema.php
```diff
@@ -29,6 +29,9 @@ public static function activate() {
 
     // Force a long client-side caching time for assets with "ver" query string.
     static::ensureAssetsCacheControl();
+
+    // Includes the necessary response headers for all requests in htaccess file.
```
sun, line 48: Code documentation should be more concise, summarizing what is being done and for what reason. "Including headers" does not tell the reader much, as headers are used for a lot of different things.
```suggestion
    // Ensures security related HTTP headers in all responses to prevent clickjacking,
    // XSS, and other vulnerabilities.
```

conf/.htaccess.headers
```diff
@@ -0,0 +1,9 @@
+# BEGIN core-standards:headers
```
sun:
```suggestion
# BEGIN core-standards:security-headers
```

conf/.htaccess.headers
```diff
@@ -0,0 +1,9 @@
...
+<IfModule mod_headers.c>
+  Header set Content-Security-Policy "frame-ancestors"
+  Header set Referrer-Policy "no-referrer-when-downgrade"
+  Header set Strict-Transport-Security "Strict-Transport-Security: max-age=31536000; includeSubDomains"
+  Header set X-Content-Type-Options "nosniff"
+  Header set X-XSS-Protection "1; mode=block"
+</IfModule>
+# END core-standards:headers
```
sun:
```suggestion
# END core-standards:security-headers
```

src/Schema.php
```diff
@@ -46,6 +49,41 @@ public static function uninstall() {
...
+      static::findAndReplaceHeader($template, 'Referrer-Policy', CORE_STANDARDS_REFERRER_POLICY);
+    }
+    if (defined('CORE_STANDARDS_STRICT_TRANSPORT_SECURITY')) {
+      static::findAndReplaceHeader($template, 'Strict-Transport-Security', CORE_STANDARDS_STRICT_TRANSPORT_SECURITY);
+    }
+    if (defined('CORE_STANDARDS_X_CONTENT_TYPE_OPTIONS')) {
+      static::findAndReplaceHeader($template, 'X-Content-Type-Options', CORE_STANDARDS_X_CONTENT_TYPE_OPTIONS);
+    }
+    if (defined('CORE_STANDARDS_X_XSS_PROTECTION')) {
+      static::findAndReplaceHeader($template, 'X-XSS-Protection', CORE_STANDARDS_X_XSS_PROTECTION);
```
sun: 1. Considering the actual usage in `wp-config.php`, it would probably be cleaner and easier to specify all of the value overrides in a single constant instead of separate constants, whose names you don't know and first have to look up?
    ```php
    const CORE_STANDARDS_X_CONTENT_TYPE_OPTIONS = '...';
    const CORE_STANDARDS_X_XSS_PROTECTION = '...';
    ```
    vs.
    ```php
    const CORE_STANDARDS_SECURITY_HEADERS = [
      'X-Content-Type-Options' => '...',
      'X-XSS-Protection' => '...',
    ];
    ```

2. What happens if a site updates to the new code and does not have any overrides configured?

3. Looking at [examples of actual production values](https://github.com/netzstrategen/nutr-nmed/blob/master/.htaccess#L194), the values in wp-config.php will actually get very (very) long.  Values in .htaccess can't use newlines without awkward workarounds.  We could make our lives easier by allowing newlines in values specified in wp-config.php and automatically transforming them into a single line (simply replacing any newlines with spaces).
    ```php
    $header_value = preg_replace('\r?\n', ' ', $header_value);
    ```

src/Schema.php
```diff
@@ -29,6 +29,9 @@ public static function activate() {
 
     // Force a long client-side caching time for assets with "ver" query string.
     static::ensureAssetsCacheControl();
+
+    // Includes the necessary response headers for all requests in htaccess file.
+    static::ensureResponseHeadersHtaccess();
```
```suggestion
    static::ensureSecurityHeaders();
```

sun (changes requested):

src/Security.php
sun: I googled "migrate x-frame-options to csp frame-ancestors" and found this:
https://stackoverflow.com/questions/43039706/replacing-x-frame-options-with-csp

And https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection states that it is deprecated and not supported by some browsers:
> it is recommended that you use [Content-Security-Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy) without allowing unsafe-inline scripts instead.

src/Schema.php
chrisbaltazar: Well that's the point, you are being too much specific of the current purpose of this new feature, when in general terms it can be used for a wider scope, meaning all kind of headers we may need in the future, besides you also had already another module/chunk called  just "security", so I'd like to keep it a bit "open to extension" here as SOLID states

src/Schema.php
chrisbaltazar: Using my new approach now there is no need to do any replacement since I'm building the whole template from the variables to reduce complexity.

src/Schema.php
chrisbaltazar: This is how the config file may look when implementing the extension of this function: 
```
const CORE_STANDARDS_HTTP_HEADERS = [
  'Strict-Transport-Security' => '"max-age=63072000; includeSubDomains; preload" env=HTTPS',
  'X-Frame-Options' => '"ALLOW-FROM https://tools.nutricia.de https://doo.net https://www.youtube.com https://vimeo.com https://games.focusgames.co.uk"',
];
```

So I think is not an issue to keep it like this for special cases only

src/Schema.php
```diff
@@ -46,6 +63,24 @@ public static function uninstall() {
     Admin::removeAccessCapability();
   }
 
+  /**
+   * Ensures dynamic response headers for all requests in htaccess.
+   */
+  public static function ensureResponseHeadersHtaccess() {
+    $headers = self::HTTP_RESPONSE_HEADERS;
+    if(defined('CORE_STANDARDS_HTTP_HEADERS') && is_array(CORE_STANDARDS_HTTP_HEADERS)) {
```
```suggestion
    if (defined('CORE_STANDARDS_HTTP_HEADERS') && is_array(CORE_STANDARDS_HTTP_HEADERS)) {
```

src/Schema.php
```diff
@@ -46,6 +63,24 @@ public static function uninstall() {
...
+      $headers = array_merge($headers, CORE_STANDARDS_HTTP_HEADERS);
+    }
+    $template = '# BEGIN core-standards:headers' . PHP_EOL;
+    $template .= '<IfModule mod_headers.c>' . PHP_EOL;
+    foreach ($headers as $header => $value) {
+      $template .= "\tHeader set $header $value" . PHP_EOL;
+    }
+    $template .= '</IfModule>' . PHP_EOL;
+    $template .= '# END core-standards:headers' . PHP_EOL;
```
sun: All files are using Unix line endings, also on Windows. Therefore, we do not use PHP_EOL, but simply "\n". Using a literal line break (by pressing the enter key) is frequently also a good option.

Also, we use two spaces instead of tabs also in .htaccess.
```suggestion
    $template = "# BEGIN core-standards:headers\n";
    $template .= "<IfModule mod_headers.c>\n";
    foreach ($headers as $header => $value) {
      $template .= "  Header set $header $value\n";
    }
    $template .= "</IfModule>\n";
    $template .= "# END core-standards:headers\n";
```

src/Schema.php
```diff
@@ -12,6 +12,20 @@
  */
 class Schema {
 
+  /**
+   * Default http headers useful to prevent clickjacking issues.
+   * Please note how the value of each header must include the quotes when needed.
+   * In case any specific project wants to extend/override these from the config/env
+   * Using the constant called CORE_STANDARDS_HTTP_HEADERS.
```
sun: - PHPDoc requires a blank PHPDoc line between the summary line and the description.
- HTTP is an acronym; acronyms should always be written all-uppercase.

```suggestion
   * HTTP headers to prevent clickjacking, XSS, and other vulnerabilities.
   *
   * The value for each header must include quotes where needed.
   * The defaults can be overridden by defining a constant named
   * CORE_STANDARDS_HTTP_HEADERS.
```

src/Schema.php
sun: I would not advise to mix non-security related headers into this block. If a feature comes up that needs further HTTP response headers, it can introduce a new block.

You're right; there is already a module named security and it should have been named better. It predates our current practice of organizing one feature into one class (instead of grouping them by topic/domain), to yield a better overview of what functionality is affected or introduced by a plugin.

sun (approved):
Thanks for addressing the issues! 👍  Please squash-merge using the PR title as commit message.

We should bump the major version to announce that sites updating to the new version should perform more extensive tests of their frontend functionality (and not just deploy the new version).

PR #72, opened 2022-02-25
Fixed Elementor does not load when editing page.

chrisbaltazar:
### Ticket
https://app.asana.com/0/1200981693019240/1201864475162938

### Description
Ad for now there is quite a discussion on forums about how Elementor is not able to work well with a strict and secure Content Policy, so this seems to be the only way to enable it

### DIFF
--- a/src/Schema.php
+++ b/src/Schema.php
@@ -20,7 +20,6 @@ class Schema {
    * CORE_STANDARDS_HTTP_HEADERS.
    */
   const HTTP_RESPONSE_HEADERS = [
-    'Content-Security-Policy-Report-Only' => 'frame-ancestors \'self\'',
     'Referrer-Policy' => '"no-referrer-when-downgrade"',
     'Strict-Transport-Security' => '"Strict-Transport-Security: max-age=31536000; includeSubDomains"',
     'X-Content-Type-Options' => '"nosniff"',
--- a/src/Schema.php
+++ b/src/Schema.php
@@ -20,7 +20,7 @@ class Schema {
    * CORE_STANDARDS_HTTP_HEADERS.
    */
   const HTTP_RESPONSE_HEADERS = [
-    'Content-Security-Policy' => '"frame-ancestors"',
+    'Content-Security-Policy-Report-Only' => '"script-src \'unsafe-inline\' \'unsafe-eval\' http: https:"',
     'Referrer-Policy' => '"no-referrer-when-downgrade"',
     'Strict-Transport-Security' => '"Strict-Transport-Security: max-age=31536000; includeSubDomains"',
     'X-Content-Type-Options' => '"nosniff"',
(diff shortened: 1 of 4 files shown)


src/Schema.php
```diff
@@ -20,7 +20,7 @@ class Schema {
    * CORE_STANDARDS_HTTP_HEADERS.
    */
   const HTTP_RESPONSE_HEADERS = [
-    'Content-Security-Policy' => '"frame-ancestors"',
+    'Content-Security-Policy-Report-Only' => '"script-src \'unsafe-inline\' \'unsafe-eval\' http: https:"',
```
sun: I don't really understand why we need to add `script-src` now.  If none is specified, then the browser should not apply any restrictions for scripts.

Did you try this?
```suggestion
    'Content-Security-Policy' => '"frame-ancestors \'self\'"',
```

sun: Seems to work correctly on https://qa.bega.wohn.netz.rocks/wp-admin/post.php?post=445166&action=elementor

chrisbaltazar: yes seem to work here as well https://stage.gaco.netz.rocks/wp-admin/post.php?post=76422&action=elementor, although I changed it again for now to work only as a reporter and dig in a a bit more in a separate task as you mention

src/Schema.php
```diff
@@ -20,7 +20,7 @@ class Schema {
    * CORE_STANDARDS_HTTP_HEADERS.
    */
   const HTTP_RESPONSE_HEADERS = [
-    'Content-Security-Policy' => '"frame-ancestors"',
+    'Content-Security-Policy-Report-Only' => 'frame-ancestors \'self\'',
```
sun: The wrapping double quotes are missing.
```suggestion
    'Content-Security-Policy-Report-Only' => '"frame-ancestors \'self\'"',
```

However, the long-term consequence of adding this line will be that a site cannot remove it or replace it with an actual CSP (without "-Report-Only").  Due to the array_merge(), the defaults are always applied.  I don't know what happens if both a CSP and a CSPRO are sent, and I don't think it makes sense to find out.

Yes, we could add support for setting NULL values as a way to "unset" the CSPRO.  But at that point I would say the configuration of our solution would borderline become more complicated than the headers themselves.

As a result, none of the solutions seem mature enough to be distributed to all of our projects currently.

Based on these grounds, I would advise to move forward without the CSP/CSPRO at all for now:
```suggestion
```

Instead, we should create a follow-up task to investigate whether we can automatically generate a safe(r) Content Security Policy for all of our sites — or whether we need to introduce a different way to generate it.

(With regard to different ways, for example, we could have a crawler that checks all pages of a website to collect all the domains that are used by fonts, scripts, styles, media, and frames, so we can review them and compose a corresponding CSP.)

sun (changes requested):

src/Schema.php
chrisbaltazar: Alright, removed now

chrisbaltazar:
Task ready to be addressed when possible https://app.asana.com/0/809933051638353/1201905380151027

sun (approved):

PR #74, opened 2022-08-02
Added delete orphans post meta cron event.

creative-andrew:
### Ticket
https://app.asana.com/0/0/1202584852391692/f

### DIFF
--- a/core-standards.php
+++ b/core-standards.php
@@ -2,7 +2,7 @@
 
 /*
   Plugin Name: Core Standards
-  Version: 3.0.2
+  Version: 3.0.3
   Text Domain: core-standards
   Description: Standard refinements.
   Author: netzstrategen
--- a/src/Admin.php
+++ b/src/Admin.php
@@ -27,11 +27,11 @@ class Admin {
   const CRON_EVENT_REVISION_CLEANUP_RETAIN_DAYS = 90;
 
   /**
-   * Cron event name for delete orphan post meta.
+   * Cron event name for delete orphan postmeta.
    *
    * @var string
    */
-  const CRON_EVENT_DELETE_ORPHAN_POST_META = Plugin::PREFIX . '/cron/delete-orphan-post-meta';
+  const CRON_EVENT_POSTMETA_CLEANUP = Plugin::PREFIX . '/cron/postmeta-cleanup';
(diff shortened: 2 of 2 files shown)


src/Admin.php
```diff
@@ -342,6 +355,14 @@ public static function cron_revision_cleanup($limit = 100) {
     }
   }
 
+  /**
+   * Cron event callback to delete orphan post meta.
+   */
+  public static function cron_delete_orphan_post_meta() {
+    global $wpdb;
+    $wpdb->query($wpdb->prepare("DELETE pm FROM {$wpdb->prefix}postmeta pm LEFT JOIN {$wpdb->prefix}posts p ON p.ID = pm.post_id WHERE p.ID IS NULL"));
```
creative-andrew: Not 100% sure if we need prepare here as we are not using any variables.

sun: Can be removed, yeah.

src/Admin.php
```diff
@@ -26,6 +26,13 @@ class Admin {
    */
   const CRON_EVENT_REVISION_CLEANUP_RETAIN_DAYS = 90;
 
+  /**
+   * Cron event name for delete orphan post meta.
+   *
+   * @var string
+   */
+  const CRON_EVENT_DELETE_ORPHAN_POST_META = Plugin::PREFIX . '/cron/delete-orphan-post-meta';
```
sun: Let's make this consistent with the existing constant name, i.e. `CRON_EVENT_POSTMETA_CLEANUP`

We can rename the value accordingly.

sun (approved):

core-standards.php
```diff
@@ -2,7 +2,7 @@
 
 /*
   Plugin Name: Core Standards
-  Version: 3.0.2
+  Version: 3.0.3
```
sun, line 5: @geist-ahnen This should have been a minor version bump, because it is a new feature and not a bugfix.

Also keep in mind that the version needs to be bumped in the main plugin file and in `package.json`.

But it's fine now, no need to do anything.

creative-andrew: Ok thanks for mentioning. I wasn't aware. I will keep in mind for the future.

PR #76, opened 2022-08-10
Fixed Elasticsearch index not in sync after imports.

creative-andrew:
### Ticket
https://app.asana.com/0/1141864033797921/1202621388598985/f

I've tested it with a local installation of ElasticSearch,Kibana and ElasticPress, doing modifications of the xml file after the imports and querying the index to see if it has the latest results. 

Works as expected.


src/Plugin.php
```diff
@@ -107,6 +107,9 @@ public static function init() {
     add_filter('script_loader_tag', __NAMESPACE__ . '\Asset::loader_tag', 10, 3);
     add_filter('style_loader_tag', __NAMESPACE__ . '\Asset::loader_tag', 10, 3);
 
+    // Filter whether to kill elasticpress sync during imports.
+    add_filter('ep_sync_indexable_kill', '__return_false', 1, 2);
```
sun: Why accepting two arguments?

```suggestion
    add_filter('ep_sync_indexable_kill', '__return_false', 1);
```

src/Plugin.php
```diff
@@ -107,6 +107,9 @@ public static function init() {
     add_filter('script_loader_tag', __NAMESPACE__ . '\Asset::loader_tag', 10, 3);
     add_filter('style_loader_tag', __NAMESPACE__ . '\Asset::loader_tag', 10, 3);
 
+    // Filter whether to kill elasticpress sync during imports.
```
```suggestion
    // Update Elasticsearch index during imports.
```

sun (approved):
Thank you! 👍  Please squash-merge using the PR title as commit message.

src/Plugin.php
creative-andrew: It passes 2 arguments right? http://10up.github.io/ElasticPress/ep_sync_indexable_kill.html

src/Plugin.php
creative-andrew: I don't think that comment makes sense with the filter.
If I see  `// Update Elasticsearch index during imports.` and something returning false I would expect that it doesn't get updated.

src/Plugin.php
sun: Yes, but your callback is accepting zero, so it is unnecessary to register a specific amount.

src/Plugin.php
sun: The comment does not need to repeat the double negation of the filter hook and its value.  Instead, the docs should explain as short and simple as possible what the customization is doing.

But if you prefer something else, maybe this works for you:

```suggestion
    // Prevent Elasticsearch becoming out of sync after importing data.
```

sun (approved):
See my replies above.

src/Plugin.php
creative-andrew: Thanks for clarifying

PR #77, opened 2022-08-23
Added caching for date filter options on administrative listing pages.

sun:
Task
- [Slow query Real Media Library Assets](https://app.asana.com/0/0/1202845994899013)

Code inspired by Automattic's WordPress.com VIP mu-plugin:
- https://github.com/Automattic/vip-go-mu-plugins-built/blob/master/performance/vip-tweaks.php

Related issues
- https://core.trac.wordpress.org/ticket/41675
- https://wordpress.stackexchange.com/questions/200376/disable-slow-media-queries

Notes
- This does not do anything on sites that do not have a persistent object cache set up.

TBD
- Should this use transients instead to also work on smaller sites that do not have a lot of infrastructure?

### DIFF
--- a/src/AdminDateFilterCache.php
+++ b/src/AdminDateFilterCache.php
@@ -1,0 +1,104 @@
+<?php
+
+/**
+ * @file
+ * Contains \Netzstrategen\CoreStandards\AdminDateFilterCache.
+ *
+ * @thanks https://github.com/Automattic/vip-go-mu-plugins-built/blob/master/performance/vip-tweaks.php
+ */
+
+namespace Netzstrategen\CoreStandards;
+
+class AdminDateFilterCache {
+
+  /**
+   * Caches an expensive query for available months/years in media library filters.
+   *
+   * @implements media_library_months_with_files
+   */
+  public static function media_library_months_with_files($months) {
+    // Abort if something else is customizing the value.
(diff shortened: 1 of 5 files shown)


src/AdminDateFilterCache.php
```diff
@@ -0,0 +1,104 @@
...
+   *
+   * @return object[]
+   *
+   * @see \WP_List_Table::months_dropdown()
+   * @see wp_enqueue_media()
+   */
+  public static function getAvailableMonths($post_type) {
+    global $wpdb;
+
+    $months = wp_cache_get($post_type, 'admin_date_filter');
```
creative-andrew, line 54: Wouldn't it make sense to prefix the key as they do?  Or do you think the group is enough? 
```suggestion
    $months = wp_cache_get("available_filter_months_$post_type", 'admin_date_filter');
```

creative-andrew (approved):
LGTM.

src/AdminDateFilterCache.php
sun: Not sure, but I think the genetic post type group is fine. We're maintaining the cache item already, so there's no invalidation issue. The only possibility could be that it's then invalidated more often with the whole group, but I think if something does that then it's okay to include this.

PR #78, opened 2023-02-21
Added multilingual support for WP CLI.

fabianmarz:
### Description
- Adds support for a `--lang` argument which accepts either the locale or `all` to return results for every language (if the argument is omitted completely the default language is used)
- Since it's invoked before every command, this should work for every command invoked via CLI
- Currently supporting WPML only

💡Alternatively copy this to your functions.php or plugin.php somewhere where it gets invoked at the right time
```php
\WP_CLI::add_hook('before_run_command', function(array $args, array $assoc_args, array $options): array {
  global $sitepress;
  if ($sitepress && $lang = $assoc_args['lang'] ?? '') {
    $sitepress->switch_lang($lang, TRUE);
  }
  return $args;
});
```

### Example results
```
$ wp post list --posts_per_page=10 --lang=fr --fields=title
+-------------------------------------------------------------------------+
| post_title                                                              |
+-------------------------------------------------------------------------+
| Les ambassadeurs EASY CONTROL sont là !                                 |
| Carrefour de l'Eau - Rennes 25 et 26 janvier 2023                       |
|  Découvrez les coulisses du chantier de Mougins                         |
| XTREMFIX : Une gamme adaptée pour toutes les zones à fortes contraintes |
| Le Tour de France HAURATON                                              |
| Hauraton nettoie ta plage                                               |
| FASERFIX                                                                |
| FASERFIX Basalte                                                        |
| Pôle Projet Hauraton                                                    |
| RECYFIX PRO                                                             |
+-------------------------------------------------------------------------+

$ wp post list --posts_per_page=10 --lang=it --fields=title
+-------------------------------+
| post_title                    |
+-------------------------------+
| SIDE-LOCK                     |
| Nuovi colori Griglia FIBRETEC |
| BIM                           |
| FASERFIX SUPER Nuovo Telaio   |
| FASERFIX Basalto              |
| FASERFIX                      |
| RECYFIX                       |
| Canali di drenaggio           |
| Classi di carico              |
| Gestione acqua piovana        |
+-------------------------------+

$ wp post list --posts_per_page=10 --lang=all --fields=title
+---------------------------------------------------+
| post_title                                        |
+---------------------------------------------------+
| Les ambassadeurs EASY CONTROL sont là !           |
| Regenwassermanagement im Neubaugebiet             |
| SIDE-LOCK                                         |
| Hauraton wächst auch 2022                         |
| Carrefour de l'Eau - Rennes 25 et 26 janvier 2023 |
| News Post Example                                 |
| Nuovi colori Griglia FIBRETEC                     |
| Klimafestival                                     |
|  Découvrez les coulisses du chantier de Mougins   |
| BIM                                               |
+---------------------------------------------------+

$ wp post list --posts_per_page=10 --fields=title
+---------------------------------------+
| post_title                            |
+---------------------------------------+
| Regenwassermanagement im Neubaugebiet |
| Hauraton wächst auch 2022             |
| News Post Example                     |
| Klimafestival                         |
| Muldensanierung                       |
| Hauraton auf der GaLaBau-Messe        |
| Presse-Splitter Ausgabe 4             |
| Grünmulden                            |
| Regelwerke                            |
| Grünmulden zur Muldenversickerung     |
+---------------------------------------+

$ wp term list reference_category --lang=bg --number=10 --fields=slug
+-------------+
| slug        |
+-------------+
| 01000001-bg |
| 01000003-bg |
| 01000004-bg |
| 01000005-bg |
| 01000006-bg |
| 01000007-bg |
| 01000010-bg |
| 01000011-bg |
| 01000017-bg |
| 01000018-bg |
+-------------+

$ wp term list reference_category --lang=fr --number=10 --fields=slug
+-----------------------+
| slug                  |
+-----------------------+
| _aucune-indication-fr |
| 0-sans-indication-fr  |
| 01000001-fr           |
| 01000003-fr           |
| 01000004-fr           |
| 01000005-fr           |
| 01000006-fr           |
| 01000007-fr           |
| 01000010-fr           |
| 01000011-fr           |
+-----------------------+
```


core-standards.php
```diff
@@ -47,4 +47,5 @@ function classloader($class) {
 if (defined('WP_CLI') && WP_CLI) {
   \WP_CLI::add_command('user export-csv', __NAMESPACE__ . '\Commands\UserExport');
   \WP_CLI::add_command('user import-csv-raw', __NAMESPACE__ . '\Commands\UserImport');
+  \WP_CLI::add_hook('before_run_command', __NAMESPACE__ . '\Hooks\MultilingualSupport::changeLanguage');
```
sun: This is only affecting WP-CLI and no regular runtime functionality, so I would keep it in `Commands`.

I'd also rename it to Sitepress to implicitly clarify that it's only integrating with WPML and not other multilingual solutions.
```suggestion
  \WP_CLI::add_hook('before_run_command', __NAMESPACE__ . '\Commands\Sitepress::changeLanguage');
```

Alternatively, if you intend to support others in the future:
```suggestion
  \WP_CLI::add_hook('before_run_command', __NAMESPACE__ . '\Commands\Language::limitLanguage');
```

src/Hooks/MultilingualSupport.php
```diff
@@ -0,0 +1,30 @@
...
+ * @file
+ * Contains \Netzstrategen\CoreStandards\Hooks\MultilingualSupport.
+ */
+
+namespace Netzstrategen\CoreStandards\Hooks;
+
+class MultilingualSupport {
+
+  /**
+   * Changes the command language to return entities of matching language only.
```
sun:
```suggestion
   * Sets the language for the current process limiting results to specified language.
```

src/Hooks/MultilingualSupport.php
```diff
@@ -0,0 +1,30 @@
...
+   *
+   * @param array $args
+   * @param array $assoc_args
+   * @param array $options
+   *
+   * @return array
+   */
+  public static function changeLanguage(array $args, array $assoc_args, array $options): array {
+    global $sitepress;
+    if ($sitepress && $lang = $assoc_args['lang'] ?? '') {
```
sun: Not all commands are supporting arbitrary options and WP-CLI's command argument/synopsis validator will respond with an error due to passing an unknown option `--lang`.

To avoid this we could unset the lang option after processing it here. But then the option would never be applied to the actual command, which might be needed as well.

Maybe worth to just add as a `// @todo` remark here and we'll see how it goes.

sun (approved):

src/Hooks/MultilingualSupport.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+namespace Netzstrategen\CoreStandards\Hooks;
+
+class MultilingualSupport {
+
+  /**
+   * Changes the command language to return entities of matching language only.
+   *
+   * Currently supporting WPML only.
```
sun: Please add that either a locale or the literal string "all" can be passed.  And that the default language is applied if nothing is passed.

core-standards.php
fabianmarz: Personally I'd rather stick with the structure as is since there might be some more hooks to come in the future and also to keep it consistent. If moving it under the command namespace it won't be very clear since the namespace imply to expect something else.

I chose a more generic name in case we want to extend the support for e.g. Polylang we wouldn't have to touch this line again and only adjust the method itself. But I'll rename the class and method to your suggestion :-)

src/Hooks/MultilingualSupport.php
fabianmarz: Good point! I've never seen the validation and couldn't find anything about throwing a notice or an error. I'll just unset it since it wasn't supported before neither. I'll be rather on the safe side not having the language information within the invoked command than breaking some commands for existing installations.

PR #85, opened 2023-08-18
Fixed fast 404.

navidabdi:
### Ticket
https://app.asana.com/0/1141864033797921/1205150148931369/f

### Description
-

### DIFF
--- a/conf/.htaccess.fast404
+++ b/conf/.htaccess.fast404
@@ -3,10 +3,10 @@
   # Fast 404 error responses for missing uploaded files, skipping
   # WordPress "File not found" page rendering.
   RewriteEngine On
-  RewriteCond %{REQUEST_URI} ^/wp-content/uploads/.+$
-  RewriteCond %{REQUEST_FILENAME} !-f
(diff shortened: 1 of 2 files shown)


conf/.htaccess.fast404
```diff
@@ -3,10 +3,10 @@
   # Fast 404 error responses for missing uploaded files, skipping
   # WordPress "File not found" page rendering.
   RewriteEngine On
-  RewriteCond %{REQUEST_URI} ^(UPLOAD_DIR_REPLACE).+$
```
fabianmarz: This is a placeholder which gets replaced when actually writing the `.htaccess` since not all WP installations are using the default values this needs to be dynamic.

See: https://github.com/netzstrategen/wordpress-core-standards/blob/master/src/Schema.php#L117

fabianmarz (changes requested):

fabianmarz:
@navidabdi, I reverted the changes you did and pushed another fix. I figured out the fast404 is not working for other projects such as LAMP/GACO as well.

PR #94, opened 2026-05-18
fix: Login page not accessible after disabling plugin.

navidabdi:
## Summary
- Deactivating `core-standards` previously left a `.htaccess` block in place that returned **403 Forbidden** for `/wp-login.php` and rewrote `/login.php` to it, so the login page was unreachable until the plugin was reactivated or the rules were manually removed.
- `Schema::deactivate()` now removes every `core-standards:*` block the plugin previously wrote to the root and uploads `.htaccess` (including the legacy `core-standards:security` marker addressed by 95aefe6), restoring the site to its pre-plugin state.
- Added a marker-based `removeBlockFromFile()` helper so removal works regardless of what content an older plugin version may have written between the `# BEGIN` / `# END` markers.

## Test plan
- [x] Activate the plugin; confirm `grep -c "core-standards:" .htaccess` returns 5 and `/login.php` works while `/wp-login.php` returns 403.
- [x] Deactivate the plugin; confirm `grep -c "core-standards:" .htaccess` returns 0, uploads `.htaccess` also has 0 matches, and `/wp-login.php` loads the login form (no 403).
- [x] Reactivate the plugin; confirm the blocks are re-added and `/login.php` works again.
- [x] Upgrade-safety: with the plugin active, inject the legacy `# BEGIN core-standards:security ... # END core-standards:security` block manually, then deactivate — both legacy and current blocks should be removed.
- [x] No-op safety: deactivate on a site whose `.htaccess` is missing or already clean — no errors, no file written.


sun (changes requested):
Most of these rules are crucial for security and stability of the (production) site, we can’t remove them.

We can consider to hook their removal into `wp plugin uninstall`.

For the deactivate hook, I would suggest to only comment out the line that rejects access to the regular login front controller.  Re-enabling the plugin will revert the change.

src/Schema.php
```diff
@@ -54,13 +54,36 @@ public static function activate() {
   public static function deactivate() {
     // Remove scheduled revision cleanup cron event.
     wp_clear_scheduled_hook(Admin::CRON_EVENT_REVISION_CLEANUP);
+
+    // The plugin may be deactivated temporarily; the production security
+    // and stability rules in .htaccess must stay in place. Only the rule
+    // that blocks direct access to /wp-login.php is disabled, so the
+    // login page remains reachable while the plugin is inactive.
+    // Re-activation rewrites the security-files block from the template
+    // and naturally restores the line.
```
```suggestion
    // When temporarily disabling the plugin, disable the rule that
    // blocks direct access to /wp-login.php, so admins can log in.
    // Reenabling the plugin rewrites all templates, so the line is
    // restored.
```

sun (approved):
Thanks

